PR #5, opened 2023-09-28
CI and Code Coverage workflow

btmonier:
Adds a GitHub action to:
* Build and unit test
* Generate a code coverage report and upload codecov.io


.github/workflows/phgv2_ci.yml
```diff
@@ -0,0 +1,35 @@
+name: PHGv2 CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types:
+      - opened
+    branches:
+      - main
```
zrm22, line 9: Should we add the path part we needed for the CD branch here as well so it only triggers to source code changes?  Might be good for now, but if we include docs might need to add more to it as we may want a trigger off documentation changes as well.

zrm22 (approved):

.github/workflows/phgv2_ci.yml
btmonier: This has now been added

tcasstevens: Does this mean "opening a pull request against main"?

btmonier: @tcasstevens - I copied this from the BioKotlin repo. That is my interpretation as well

build.gradle.kts
```diff
@@ -68,7 +68,7 @@ tasks.test {
 }
 
 tasks.withType<KotlinCompile> {
-    kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions.jvmTarget = "11"
```
tcasstevens, line 71: Are you sure about this? I don't know what it should be, but the Ci deployment is compiling with java-version 17

tcasstevens (approved):

build.gradle.kts
btmonier: @tcasstevens - should we update this to 17?

tcasstevens: @btmonier If it compiles and works correctly I would say yes

PR #22, opened 2023-10-06
Unit test variant loading utils

zrm22:
## Description

This adds in the unit tests for VariantLoadingUtils.  It currently does not test the gzipping and the indexing of the files as those require 3rd party tools which will need to be installed on the CI first.  Coverage is in the 85% range.

## Type of change

- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Checklist:

## Changelog entry

```release-note
Added in more error checking for the creation of VariantContexts and added in additional unit tests.
```

### DIFF
--- a/src/test/kotlin/net/maizegenetics/phgv2/utils/VariantLoadingUtilsTest.kt
+++ b/src/test/kotlin/net/maizegenetics/phgv2/utils/VariantLoadingUtilsTest.kt
@@ -1,7 +1,9 @@
 package net.maizegenetics.phgv2.utils
 
 import biokotlin.seq.NucSeq
+import htsjdk.variant.variantcontext.VariantContext
 import htsjdk.variant.vcf.VCFAltHeaderLine
+import htsjdk.variant.vcf.VCFFileReader
 import htsjdk.variant.vcf.VCFHeaderVersion
 import org.junit.jupiter.api.Assertions
 import org.junit.jupiter.api.Assertions.assertNotEquals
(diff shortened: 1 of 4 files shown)


src/test/kotlin/net/maizegenetics/phgv2/utils/VariantLoadingUtilsTest.kt
```diff
@@ -1,22 +1,129 @@
...
+        assertEquals(expectedVariant.getAttributeAsInt("ASM_Start",0), loadedVariant.getAttributeAsInt("ASM_Start",0), "ASM_Start not as expected")
+        assertEquals(expectedVariant.getAttributeAsInt("ASM_End",0), loadedVariant.getAttributeAsInt("ASM_End",0), "ASM_End not as expected")
+
     }
 
     @Test
     fun testCreateGenericHeader() {
         // this also tests createGenericHeaderLineSet
+        val taxa = listOf("LineA", "LineB", "LineC")
+        val altHeaders = (1 .. 10).map{VCFAltHeaderLine("<ID=${it}, Description=\"Description for ${it}\">", VCFHeaderVersion.VCF4_2)}.toSet()
```
lynnjo, line 63: THis will be changed eventually to how the md5 hash as id, correct?  (I guess that is the source code that needs to be whatever we've decided goes into the alt headers)

lynnjo (approved):

src/test/kotlin/net/maizegenetics/phgv2/utils/VariantLoadingUtilsTest.kt
zrm22: Yes this will be hashes normally.  For the test it just needs to be a String though so it does still work.  Once we have the G/H VCF code up and running that can be used to test storing the hashes here.

PR #38, opened 2023-10-25
implemented additional AGC commands

lynnjo:
## Description

## Type of change

## Checklist:

## Changelog entry

```release-note
Additional AGC commands implemented
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
@@ -4,9 +4,9 @@ import biokotlin.seq.NucSeq
 import java.io.BufferedInputStream
 import java.io.ByteArrayOutputStream
 import java.io.File
-import java.util.logging.Logger
+import org.apache.logging.log4j.LogManager
 
-private val myLogger = Logger.getLogger("net.maizegenetics.phgv2.utils.SeqUtils")
+private val myLogger = LogManager.getLogger("net.maizegenetics.phgv2.utils.SeqUtils")
 
 /**
  * The file contains functions to pull data from an AGC compressed file.
@@ -96,13 +96,13 @@ fun retrieveAgcData(dbPath:String,agcCmdList:List<String>):List<String>? {
                 dataFromListSet(agcOut)
             }
             else -> {
-                myLogger.severe("Invalid AGC command sent to retrieveAgcData: $agcCmd")
+                myLogger.error("Invalid AGC command sent to retrieveAgcData: $agcCmd")
                 throw IllegalStateException("Invalid AGC command sent to retrieveAgcData: $agcCmd")
             }
         }
         return queryData
     } catch (exc: Exception) {
-        myLogger.severe("Error reading AGC output: ${exc.message}")
+        myLogger.error("Error reading AGC output: ${exc.message}")
         throw IllegalStateException("Error reading AGC output: ${exc.message}")
     } finally {
         agcOut.close()
@@ -125,7 +125,7 @@ fun dataFromListSet(agcOut:BufferedInputStream):List<String>{
             }
         }
     } catch (exc: Exception) {
-        myLogger.severe("Error reading AGC output: ${exc.message}")
+        myLogger.error("Error reading AGC output: ${exc.message}")
     } finally {
         agcOut.close()
     }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
```diff
@@ -9,36 +9,195 @@ import java.util.logging.Logger
...
+        // call appropriate function to process the output
+        val queryData = when (agcCmd) {
+            "listctg" -> {
+                dataFromListCtgs(agcOut, genomes)
+            }
+            "listset" -> {
+                dataFromListSet(agcOut)
+            }
+            else -> {
+                myLogger.severe("Invalid AGC command sent to retrieveAgcData: $agcCmd")
```
zrm22, line 99: Couldn't this just be a myLogger.error?

lynnjo: Yes, well that is the whole logger issue.  "error" doesn't show up as an option, perhaps because I'm using: (copied from somewhere initially)

private val myLogger = Logger.getLogger("net.maizegenetics.phgv2.utils.SeqUtils")

THis is from import java.util.logging.Logger

I thought that "private val myLogger = LogManager.getLogger(AgcCompress::class.java)" required a class, and the SeqUtils.kt is not a class.

I'm happy to change this however you think it can be changed.  I know the code in this repository has inconsistent logging methods.

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
```diff
@@ -9,36 +9,195 @@ import java.util.logging.Logger
...
+ * Processes AGC commands listset or listctg
+ */
+fun retrieveAgcData(dbPath:String,agcCmdList:List<String>):List<String>? {
+
+    val agcFile = "${dbPath}/assemblies.agc"
+
+    check(File(agcFile).exists()) { "File assemblies.agc does not exist in folder $dbPath- please send a valid path that indicates the parent folder for your assemblies.agc compressed file." }
+    check(agcCmdList.isNotEmpty()) { "No AGC command was sent to retrieveAgcData" }
+
+    val agcCmd = agcCmdList[0]
```
zrm22, line 71: Do we ever use more than the first element?  If not, I would consider using an enum if possible so we can force the commands to be within a controlled vocabulary

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
zrm22: If its not letting you do an error this is probably fine.  Will still throw an exception and stop processing.

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
```diff
@@ -9,36 +9,195 @@ import java.util.logging.Logger
 private val myLogger = Logger.getLogger("net.maizegenetics.phgv2.utils.SeqUtils")
 
 /**
- * The retrieveAgcContigs function is called to pull data using the "agc getctg" command.
- *  The user command list should have components that look like one of the following:
+ * The file contains functions to pull data from an AGC compressed file.
+ * It is assumed the AGC file is named "assemblies.agc" and is in the same folder as the tiledb datasets.
```
tcasstevens, line 13: Does this mean named with .agc extension? Or exactly that?

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
lynnjo: Yes, we do.  If the command is listctg (vs listset), then  additional elements indicating the genomes needs to be on the list.

listset: returns a list of genomes (sample names) in the agc file.  No additional parameters.
listctg: returns a list of contigs for the specified genomes.  Must have at least 1 genome on the list., so command plus genomes are on the list.

tcasstevens (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
lynnjo: Yes, it compiles.  It is used Logger.getLogger() is used in CreateRefVcf.kt, VariantLoadingUtils.kt and SeqUtils.kt.  The other classes in clikt are using LogManager.getLogger().  I can see about changing this

tcasstevens: That must be coming from another dependency. Our build.gradle.kts has a dependency on Log4j2

lynnjo: I changed it to use apache LogManager as with the cli classes.

pjbradbury (approved):

PR #40, opened 2023-10-27
Create fasta from hvcf

zrm22:
## Description

This is the initial Pull Request for the code to pull the sequence from the AGC record using the information within the hvcf file.

Note: Right now the unit tests are taking a long time and it is due to a slow pull of the data from the AGC record.  There is nothing I can see that should make it slow(it runs in sub 1 second for full chromosomes from the command line), but it takes about 30 seconds to get 40 haplotypes.  Will be testing stuff here while I address comments.

## Type of change

## Checklist:

## Changelog entry

```release-note
Added in create-fasta-from-hvcf command.
```


btmonier (approved):

tcasstevens (approved):

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        val ranges = rangesAndOtherInfo.map { it.first }.flatten()
+        val seqs = retrieveAgcContigs(dbPath,ranges)
+
+        return rangesAndOtherInfo.map { it.third.copy(sequence = seqs[it.second]!!.seq()) }
+    }
+
+    /**
+     * Function to output composite contig sequences to a fasta file
+     */
```
lynnjo, line 131: A quick note on what a "composite contig sequence" is would be warranted here.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -0,0 +1,180 @@
...
+import com.github.ajalt.clikt.parameters.options.validate
+import htsjdk.variant.variantcontext.VariantContext
+import htsjdk.variant.vcf.VCFFileReader
+import htsjdk.variant.vcf.VCFHeader
+import net.maizegenetics.phgv2.utils.retrieveAgcContigs
+import java.io.BufferedWriter
+import java.io.File
+import java.io.FileWriter
+
+
```
lynnjo, line 15: This class would benefit from a header comment with an overview of what the code in the class provides.  In particular, it should explain what the composite and haplotype fasta options are.   The terminology section says a "composite genome" is "inferred genome based on its composite set of alternate and reference haplotypes".    Terminology sections does not mention "haplotype fasta" though from the name it sounds like the haplotype and composite fastas are both haplotype fastas.

Some detail on why/when the user should opt for one over the other would be useful.

zrm22: I added this in, but the user facing documentation will be updated by @btmonier in the future to better describe this to the people actually using the code.

lynnjo (approved):

PR #48, opened 2023-11-09
added envFile param, code to use default env file if no param

lynnjo:
## Description

## Type of change

## Checklist:

## Changelog entry

SetupEnvironment enhanced to take user envFile, or find default environment file when none is provided.

```release-note

```


src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
```diff
@@ -39,22 +60,21 @@ class SetupEnvironment : CliktCommand(help="create conda environment for PHG") {
                 .filter { it.contains("prefix already exists") }
                 .toList()
             if (success == null || success.size == 0) {
-                println("conda env create command for file ${envFile} run via ProcessBuilder returned error code $error")
+                myLogger.severe("conda env create command for file ${envFile} run via ProcessBuilder returned error code $error")
```
tcasstevens, line 63: What does severe mean? Is that different than error?

src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
```diff
@@ -17,19 +19,38 @@ import java.util.stream.Collectors
...
     fun createEnvironment( envFile: String, outputDir:String) {
+
+        var selectedEnvFile = envFile
+        myLogger.info("begin run: selectedEnvFile = $selectedEnvFile")
+
+        // if no user defined environment file, use the default
+        if (selectedEnvFile == "") {
+
+            selectedEnvFile = "${outputDir}/environment.yml"
+            val inputUrl = SetupEnvironment::class.java.getResource("environment.yml")
```
tcasstevens, line 38: Is this doing the same thing as line 41?

src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
```diff
@@ -17,19 +19,38 @@ import java.util.stream.Collectors
  * creating it, do we assume it exists?
  */
 class SetupEnvironment : CliktCommand(help="create conda environment for PHG") {
-    // The output directory will store the log and error files from the ProcessBUilder conda create command
-    val outputDir by option("-o", "--outputDir", help = "Directory where ProcessBuilder, conda create env command log files will be written")
-        .required()
+
+    private val myLogger = Logger.getLogger("net.maizegenetics.phgv2.cli.SetupEnvironment")
+
+    val envFile by option("-e", "--env-file", help = "File containing the conda environment definition")
```
tcasstevens, line 25: I'm not sure why this is an option?

src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
lynnjo: AH, yes!   Remnants of previous attempts to get this to work.  I'll delete line 38, it isn't needed.

src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
lynnjo: @tcasstevens how should this be coded as an optional parameter?

tcasstevens: I would have thought not having this parameter at all?

lynnjo: We decided to allow them to use their own environment file in case they for some reason needed something additional.  Perhaps a conda environment for 2 things?  @zrm22 what are your thoughts?

src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
lynnjo: For the logger used, there is no "error".  "Severe" is the only option above warning.  If I'm using the wrong logger, let me know and I'll change that.

src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
zrm22: I think it is helpful for the user to submit their own.  As lynn mentioned we might not have some software installed in our environment that a user may want to include.  By allowing them to submit their own yaml file we can accommodate this.  Most people likely will just do the conda commands themselves though for external env, but it doesn't hurt to allow users to do it within our system.

src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
tcasstevens: You're using the wrong logger

import org.apache.logging.log4j.LogManager
import org.apache.logging.log4j.Logger

tcasstevens (approved):

src/main/kotlin/net/maizegenetics/phgv2/cli/SetupEnvironment.kt
zrm22: I wonder if there is a way to get the GitHub action release task to copy this file out and into the root directory so its in both the pig_v2.jar and in the working directory for the executable?  Might be worth investigating in the future.

zrm22 (approved):

PR #53, opened 2023-11-14
Initial logging utils

tcasstevens:
## Description

## Type of change

## Checklist:

## Changelog entry

```release-note

```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/utils/LoggingUtils.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/utils/LoggingUtils.kt
@@ -1,0 +1,43 @@
+package net.maizegenetics.phgv2.utils
+
+import org.apache.logging.log4j.Level
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.core.LoggerContext
+import org.apache.logging.log4j.core.config.builder.api.ConfigurationBuilderFactory
+
+private val myOriginalOutputStream = System.out
+private val myOriginalErrStream = System.err
+
+/**
+ * Sends all logging messages (including log4j) to standard out. The logging
+ * level will be DEBUG.
+ */
+fun setupDebugLogging() {
+    System.setOut(myOriginalOutputStream)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/maizegenetics/phgv2/utils/LoggingUtils.kt
```diff
@@ -0,0 +1,43 @@
+package net.maizegenetics.phgv2.utils
+
+import org.apache.logging.log4j.Level
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.core.LoggerContext
+import org.apache.logging.log4j.core.config.builder.api.ConfigurationBuilderFactory
+
+private val myOriginalOutputStream = System.out
+private val myOriginalErrStream = System.err
+
```
lynnjo, line 10: When do these get called?  Do each of the classes/files need to include a call to one of these functions?

tcasstevens: That's a good question. In TASSEL, I had the GUI and command line call it automatically, so that everything else didn't need to. But yeah, seems like each Clikt command will need to call the setup method. I wonder if we can put it in a central place so that units tests will automatically get it. I need to think about this more.

lynnjo (approved):

pjbradbury (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/LoggingUtils.kt
zrm22: Could it maybe go in PHG.kt? It looks like there is an init block which might be able to be used?

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/LoggingUtils.kt
tcasstevens: @zrm22 Why didn't I think of that! Yes, done

PR #57, opened 2023-11-17
Handle inversions

zrm22:
## Description

Changing how we handle inverted coordinates in HVCF ALT tags.

## Type of change

## Checklist:

## Changelog entry

```release-note
Changing how assembly coordinates are stored in HVCF. Now we keep track of all sub-sequences.
```


src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
```diff
@@ -369,13 +482,13 @@ class CreateMafVcf : CliktCommand(help = "Create gVCF and hVCF from Anchorwave M
...
-                            "Number=6,Source=\"${dbPath}/assemblies.agc\"," +
-                            "Contig=\"${metaDataRecord.asmContig}\",Start=\"${metaDataRecord.asmStart}\"," +
-                            "End=\"${metaDataRecord.asmEnd}\",Checksum=\"Md5\",RefRange=\"${refSeqHash}\">",
-                    VCFHeaderVersion.VCF4_2
-                )
+            VCFAltHeaderLine(
+                "<ID=${assemblyHaplotypeHash}, Description=\"haplotype data for line: ${metaDataRecord.sampleName}\">," +
+                        "Number=4,Source=\"${dbPath}/assemblies.agc\"," +
+                        "Regions=\"${metaDataRecord.asmRegions.map { "${it.first.contig}:${it.first.position}-${it.second.position}" }.joinToString(",")}\"," +
+                        "Checksum=\"Md5\",RefRange=\"${refSeqHash}\">",
```
lynnjo, line 489: Is this a good idea to use a comma as a delimiter for the regions? Comma is the delimiter for fields in the ALT header.  This could have implications for parsing.  I'd like to see something else used as delimiter for regions.  You have colon and dash inside each region.  Would semi-colon work between regions?

lynnjo: @zrm22 Actually, the comma is fine.   What we have wrong is that these fields should be quoted.  From the VCF 4.3 spec under section 1.4 Meta-information lines:

##INFO=<ID=ALLELEID,Number=A,Type=String,Description="Allele ID",Source="ClinVar",Version="20220804">
In the above example, the extra fields of “Source” and “Version” are provided. **The values of optional fields must be
written as quoted strings, even for numeric values.** Other structured lines not defined by this specification may also
be used; the only default field for such lines is the required ID field.

We need to put double quotes around all the fields we have added.

https://samtools.github.io/hts-specs/VCFv4.3.pdf

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
```diff
@@ -321,30 +322,142 @@ class CreateMafVcf : CliktCommand(help = "Create gVCF and hVCF from Anchorwave M
...
+        val variantsConverted = variants.map { convertVariantContextToPositionRange(it) }
+
+        //resize the first and last position based on the strand
+        val resizedFirst = resizePositionRange(variantsConverted.first(),newStart,true)
+        val resizedLast = resizePositionRange(variantsConverted.last(),newEnd,false)
+
+        //merge the first and last with the rest of the variants
+        val mergedVariants = mutableListOf<Pair<Position,Position>>()
+        mergedVariants.add(resizedFirst)
+        mergedVariants.addAll(variantsConverted.subList(1,variantsConverted.size-1))
```
lynnjo, line 349: Shouldn't this be variantsConverted.subList(0,variantsConverted.size-1)?  When the list only has 1 element, you get an out of range exception as the range goes from 1 to 0.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
```diff
@@ -321,30 +322,142 @@ class CreateMafVcf : CliktCommand(help = "Create gVCF and hVCF from Anchorwave M
             else lastVariant.getAttributeAsInt("ASM_Start",region.second.position)
         }
 
+        val regions = buildNewAssemblyRegions(newASMStart,newASMEnd,variants)
```
lynnjo, line 325: Was the intent to remove the check for strand match on line 311 (based on the rework to use regions)?  It still fails on that line in the at-scale testing.  I'll comment that out and see how it does.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
lynnjo: These comments stemmed from looking at the created Ref.h.vcf which is now created and has lines that look like:
##ALT=<ID=06ae4e937668d301e325d43725a38c3f,Description="haplotype data for line: Ref",Number=4,Source="data/test/smallseq/Ref.fa",Regions=1:45001-49500,Checksum=Md5,RefRange=06ae4e937668d301e325d43725a38c3f>

Note that while the Description and Source field values are surrounded by double quotes, the Source, Regions, Checksum and RefRange fields are not.

However, in the code, the double quotes exist.  And I found when testing at scale, the assembly h.vcf files had double quotes around the Regions values when they were comprised of more than a single region.  See phg_v2 slack channel 11/20/23 for discussion.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
```diff
@@ -118,14 +147,23 @@ class CreateMafVcf : CliktCommand(help = "Create gVCF and hVCF from Anchorwave M
 
     /**
      * Simple Wrapper Function to call biokotlin's MAFToGVCF.getVariantContextsfromMAF
+     * The parameters to BioKotlin's (version 0.10) getVariantContextsfromMAF are:
+     * 1. mafFileName - the name of the MAF file
```
pjbradbury, line 151: Should variable names be enclosed in square brackets to create a link in the generated docs?

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -162,15 +173,47 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a fasta file from a hvc
...
+    fun buildHapSeq(seqs: Map<String,NucSeq> ,displayRegions : List<String>, hapSeqObjects: HaplotypeSequence) : String {
+        val hapSeqRegions = hapSeqObjects.asmRegions
+
+        return displayRegions.mapIndexed{ idx, currentDisplayRegion ->
+            val currentHapSeqRegion = hapSeqRegions[idx]
+            val seq = seqs[currentDisplayRegion]!!
+
+            //Means it is the first region
+            if(currentHapSeqRegion.first.position > currentHapSeqRegion.second.position) {
+                //Means it needs to be reverse complemented
```
btmonier, line 209: Similar to Peter's comment above - can you add more commentary here on what "it" is?

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -162,15 +173,47 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a fasta file from a hvc
...
             val altMetaData = altHeaders[hapId]
             //Need to subtract 1 from start as it uses 0 based format
-            val range = "${altMetaData!!.contig}@${sampleName}:${altMetaData!!.start-1}-${altMetaData!!.end-1}"
-            val outputDisplayName = "${altMetaData!!.contig}:${altMetaData!!.start-1}-${altMetaData!!.end-1}"
-            Triple(range, outputDisplayName, HaplotypeSequence(hapId, "", altMetaData.refRange, it.contig, it.start, it.end, altMetaData!!.contig, altMetaData!!.start, altMetaData!!.end))
+            val regions =  altMetaData!!.regions
+            val queryRanges = mutableListOf<String>()
+            val displayRanges = mutableListOf<String>()
+            for(region in regions) {
+                queryRanges.add("${region.first.contig}@${sampleName}:${region.first.position-1}-${region.second.position-1}")
```
btmonier, line 180: Should this be 0-based inclusive start, exclusive end? Here, you are subtracting from both start and end.

zrm22: No It needs to be 0-based inclusive inclusive.  Early on we convert the BED records to VCF coordinates(1-based inclusive inclusive), but we need to convert it here to what AGC is looking for.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -142,17 +144,26 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a fasta file from a hvc
                 //These are optional header fields so we check these in the unit test.
```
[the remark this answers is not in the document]
zrm22: I agree, I made a new trello card for this as it will take a bit of work to make sure the types all match up and the validation that happens currently will still happen.  Should be a relatively minor change.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
zrm22: I updated the comment here and above it should be more clear now.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -124,15 +125,54 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a fasta file from a hvc
             check(altHeaders.containsKey(hapId)) { "Haplotype ID $hapId not found in ALT Header" }
             val altMetaData = altHeaders[hapId]
```
lynnjo, line 126: Hvcf are always created by phg_v2, so there is no chance that altMetaData could be null, correct?  (unless the file was "tampered" with?)

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -15,7 +17,7 @@ import java.io.FileWriter
 
 
 data class HaplotypeSequence(val id: String, val sequence: String, val refRangeId: String, val refContig: String, val refStart: Int, val refEnd: Int,
```
tcasstevens, line 19: Should this have a checksum?

zrm22: The checksum is the id.  So it does not need it as it would just be duplicated.

data/test/smallseq/LineA_old_BadALTHeader.h.vcf
```diff
@@ -0,0 +1,93 @@
+##fileformat=VCFv4.2
```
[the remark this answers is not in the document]
tcasstevens: I mean in TestExtension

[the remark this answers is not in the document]
zrm22: Gotcha, no but I can quickly.

lynnjo (approved):

tcasstevens (approved):

PR #59, opened 2023-11-27
Tiledb vcf java interface

pjbradbury:
## Description

## Type of change

## Checklist:

## Changelog entry

```release-note

```


src/test/kotlin/net/maizegenetics/phgv2/cli/TestExtension.kt
```diff
@@ -19,7 +19,7 @@ class TestExtension : BeforeAllCallback {
         const val smallSeqInputDir = "data/test/smallseq/"
         const val smallseqLineAFile = "${smallSeqInputDir}LineA.fa"
         const val smallseqLineAMafFile = "${smallSeqInputDir}LineA.maf"
-        const val smallseqLineBFile = "${smallSeqInputDir}LineB.fa"
```
tcasstevens, line 22: Don't we need a constant for the fasta file?

src/main/kotlin/net/maizegenetics/phgv2/api/TileDBVcfReader.kt
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    private fun decodeVarlenChar(dataBuffer: ByteBuffer,
+                         offsetBuffer: ByteBuffer,
+                         numberOfRecords: Int,
+                         outputList: MutableList<String>)  {
+        val offsets = IntArray(numberOfRecords + 1)
+        offsetBuffer.asIntBuffer().get(offsets)
+
+        dataBuffer.rewind()
```
lynnjo, line 279: Why is rewind() needed here but not elsewhere?  I see we call decodeVarlenChar() with dataBuffer equal to:

  dbReader.getBuffer("sample_name"). or
   dbReader.getBuffer("contig").  or
   dbReader.getBuffer("alleles") (from the decodeVarlenListChar() call)
   
   In each of these cases I didn't see the databuffer being read before the decodeVaLenChar, but perhaps I missed where that happens

src/main/kotlin/net/maizegenetics/phgv2/api/TileDBVcfReader.kt
```diff
@@ -0,0 +1,344 @@
...
+        for (ndx in 0..<numberOfRecords) {
+            val sizeOfName = offsets[ndx + 1] - offsets[ndx]
```
lynnjo, line 281: Thank you for all the 3 char index variable names!  (ndx)

lynnjo (approved):

src/main/kotlin/net/maizegenetics/phgv2/api/TileDBVcfReader.kt
```diff
@@ -0,0 +1,344 @@
...
+
+        // alloc buffers
+        var bufferSize = 1000000
+        query.setDataBuffer("header", NativeArray(context, bufferSize, Datatype.TILEDB_STRING_ASCII))
+        query.setOffsetsBuffer("header", NativeArray(context, 1024, Datatype.TILEDB_UINT64))
+        var data: ByteArray
+
+        // submit query
+        do {
+            query.submit()
```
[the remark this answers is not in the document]
pjbradbury: yes. query.submit() loads data into the buffers. If bufferSize is too small for the amount of data, then the QueryStatus is incomplete. In that case, the buffers need to be processed, then query.submit run again. I could make bufferSize a user settable parameter to provide some control over memory use. I will add a comment explaining that.

src/main/kotlin/net/maizegenetics/phgv2/api/TileDBVcfReader.kt
pjbradbury: Rewind sets the data buffer pointer to the start of the buffer. If the buffer is smaller than the data, then after the buffers are processed, query.submit is run again. If there is no buffer.rewind, the pointer stays where it was after the last loop. Since this method does not know if it has been run before, its best to use rewind just in case. All the other buffer reading methods use the asIntBuffer method before retrieving data. That returns an IntBuffer set at the beginning, so rewinding is not needed.

src/test/kotlin/net/maizegenetics/phgv2/cli/TestExtension.kt
pjbradbury: If you mean LineA.fa, thats three lines up (smallseqLineAFile).

src/main/kotlin/net/maizegenetics/phgv2/api/TileDBVcfReader.kt
lynnjo: Thanks for the explanation.

tcasstevens (approved):

PR #65, opened 2023-12-07
Haplotypegraph

tcasstevens:
## Description

## Type of change

## Checklist:

## Changelog entry

```release-note

```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
@@ -85,12 +85,12 @@ class HaplotypeGraph(hvcfFiles: List<String>) {
     /**
      * Returns the hapId for the sample in the specified ReferenceRange.
      */
-    fun sampleToHapId(range: ReferenceRange, sample: String, gameteId: Int = 0): String {
+    fun sampleToHapId(range: ReferenceRange, sample: SampleGamete): String {
         val rangeId = refRangeMap[range]
         require(rangeId != null) { "sampleToHapId: range: $range not found" }
-        val sampleId = sampleNameToIdMap[sample]
+        val sampleId = sampleNameToIdMap[sample.name]
         require(sampleId != null) { "sampleToHapId: sample: $sample not found" }
-        return rangeByGameteIdToHapid[rangeId][sampleId][gameteId]
+        return rangeByGameteIdToHapid[rangeId][sampleId][sample.gameteId]
     }
 
     /**
--- a/src/test/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraphTest.kt
+++ b/src/test/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraphTest.kt
@@ -62,28 +62,28 @@ class HaplotypeGraphTest {
 
         // tests for sampleToHapId() method
 
-        var checksum = graph.sampleToHapId(ranges[0], "LineA")
+        var checksum = graph.sampleToHapId(ranges[0], SampleGamete("LineA"))
         assertEquals(
             "12f0cec9102e84a161866e37072443b7",
             checksum,
             "sampleToHapId: checksum not 12f0cec9102e84a161866e37072443b7: $checksum"
         )
 
-        checksum = graph.sampleToHapId(ranges[0], "LineB")
+        checksum = graph.sampleToHapId(ranges[0], SampleGamete("LineB"))
         assertEquals(
             "4fc7b8af32ddd74e07cb49d147ef1938",
             checksum,
             "sampleToHapId: checksum not 4fc7b8af32ddd74e07cb49d147ef1938: $checksum"
         )
 
-        checksum = graph.sampleToHapId(ranges[ranges.size - 1], "LineA")
+        checksum = graph.sampleToHapId(ranges[ranges.size - 1], SampleGamete("LineA"))
         assertEquals(
             "0eb9029f3896313aebc69c8489923141",
             checksum,
             "sampleToHapId: checksum not 0eb9029f3896313aebc69c8489923141: $checksum"
         )
 
-        checksum = graph.sampleToHapId(ranges[ranges.size - 1], "LineB")
+        checksum = graph.sampleToHapId(ranges[ranges.size - 1], SampleGamete("LineB"))
         assertEquals(
             "5031218d4ac709dd51a946acd0550356",
             checksum,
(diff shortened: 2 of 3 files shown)


src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
```diff
@@ -209,15 +220,14 @@ class HaplotypeGraph(hvcfFiles: List<String>) {
      */
     private suspend fun addSites() {
 
-        // rangeIdToSampleToChecksum[refRangeId][sampleId] -> Checksum / hapid
-        val rangeIdToSampleToChecksum = mutableListOf<Array<String?>>()
+        // rangeIdToSampleToChecksum[refRangeId][sampleId][gameteId] -> Checksum / hapid
+        val rangeIdToSampleToChecksum = mutableListOf<Array<MutableList<String?>>>()
```
zrm22, line 224: Is there a way to get away from nullable Strings?  Can we use empty strings instead?

src/test/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraphTest.kt
```diff
@@ -45,8 +44,12 @@ class HaplotypeGraphTest {
             }
         }
 
-        Sizeof.printMemoryUse()
+    }
 
+    @Test
+    fun testBadInputHaplotypeGraph() {
+        val graph = HaplotypeGraph(listOf(TestExtension.smallseqLineAHvcfFileBadAltTag))
+        check(graph == null || graph.numberOfRanges() == 0) { "graph not null or empty" }
```
zrm22, line 52: You should use an assert here instead of just a check.

lynnjo (approved):

src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
tcasstevens: Ok, I changed HaplotypeGraph.rangeByGameteIdToHapid to use "" instead of nulls. Some intermediate variables are still using null but not exposed to the user of the class.

src/test/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraphTest.kt
tcasstevens: Ok, changed this

src/test/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraphTest.kt
```diff
@@ -45,8 +44,12 @@ class HaplotypeGraphTest {
             }
         }
 
-        Sizeof.printMemoryUse()
+    }
 
+    @Test
```
pjbradbury, line 65: This now needs a test with a diploid hvcf.

src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
```diff
@@ -233,16 +243,38 @@ class HaplotypeGraph(hvcfFiles: List<String>) {
...
+        }.toTypedArray()
 
     }
 
-    private fun mergeStringArrays(seqHash1: Array<String?>?, seqHash2: Array<String?>): Array<String?> {
-        return seqHash1?.let {
-            Array(seqHash2.size) { index ->
-                if (seqHash1[index] != null) seqHash1[index] else seqHash2[index]
+    // sampleGameteHapid[sampleId][gameteId] -> Checksum / hapid
+    private fun mergeStringArrays(
```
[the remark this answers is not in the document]
tcasstevens: Ok, I added this

src/test/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraphTest.kt
tcasstevens: Do we have input for that? I thought we decided to get the haploid working first. Then make sure diploid works after push to release in December?

src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
```diff
@@ -77,25 +72,32 @@ class HaplotypeGraph(hvcfFiles: List<String>) {
      * Returns a hapId -> sample list map for the given ReferenceRange.
      * Returned Map<hapId, List<sampleName>>
      */
-    fun hapIdToSamples(range: ReferenceRange): Map<String, List<String>> {
+    fun hapIdToSamples(range: ReferenceRange, gameteId: Int = 0): Map<String, List<String>> {
```
pjbradbury, line 75: I do not think this will work very well. Since the user has to know how many gametes per sample and it is possible to construct a graph with some haploid and some diploid samples. If this is called with gamete = 1 and some hapIdList in line 80 only have a single item, then I think that generates an error. I think what is needed is to have a method that takes ReferenceRange as the only argument and returns a map of hapid -> List<Pair<sampleName, gameteId>>. There may be a better return data structure but that is how I see it.

src/test/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraphTest.kt
pjbradbury: It will be easy enough to modify an existing test hvcf. The test hvcf should include missing allele calls (e.g., A/., ./.). But, it can wait to be added later and does not have to part of this PR.

src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
zrm22: I agree with Peter.  We need to couple the sampleName and gameteIds and propagate them through like they would be the 'Taxon' in phgv1.  Alternatively to Peters suggestion we could make a simple data class which would hold the sampleName and the gameteId together.  A pair works just as well though, but a class would be more descriptive when using it.

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
tcasstevens: Sounds like a data class is needed. Should I do that now? Or are we moving forward with what I have?

btmonier (approved):

PR #67, opened 2023-12-08
Do not merge this! Testing Actions for automatic Conda build + upload

matthewwiese:
Source: https://stackoverflow.com/a/76439748

I'm working on getting automatic Conda package builds working. As part of that I want to make sure I'm setting some repository variables correctly with my GitHub Actions.

However, GitHub will not show any action in the web UI that isn't already on `main`. Triggering my action via a PR is a quick workaround so that it appears and remains on the website for triggering manually via `workflow_dispatch`.

### DIFF
--- a/.github/workflows/test_set_variable.yml
+++ b/.github/workflows/test_set_variable.yml
@@ -1,0 +1,28 @@
+name: Test Set Variable
+
+on:
+  workflow_dispatch:
+
+jobs:
+  test-set-variable:
+    runs-on: ubuntu-latest
+    steps:
+      # Update repository variables for Conda package
+      - name: Fake data for testing
+        run: |
+          export VERSION=1.2.3.4
+          export RELEASE=1.2
+          touch PHGv2-v${{ env.RELEASE }}.tar
+      - uses: action-pack/set-variable@v1
+        with:
+          name: 'PHG2_VERSION'
+          value: '${{ env.VERSION }}'
+          token: ${{secrets.PHGV2CD}}
+      - name: Calculate MD5 of package
+        run: |
+          export PHG2_VERSION_MD5=$(md5sum PHGv2-v${{ env.RELEASE }}.tar)
+      - uses: action-pack/set-variable@v1
+        with:
+          name: 'PHG2_VERSION_MD5'
+          value: '${{ env.PHG2_VERSION_MD5 }}'
+          token: ${{secrets.PHGV2CD}}
--- a/.github/workflows/test_set_variable.yml
+++ b/.github/workflows/test_set_variable.yml
@@ -1,7 +1,11 @@
 name: Test Set Variable
 
+# Hacky solution: https://stackoverflow.com/a/76439748
+# Annoying that I have to do this for GitHub to show any action not on main
+
 on:
   workflow_dispatch:
+  pull_request:
 
 jobs:
   test-set-variable:


conda/meta.yaml
```diff
@@ -0,0 +1,31 @@
+{% set name = "PHG2" %}
+
+package:
+  name: {{ name|lower }}
+  version: "{{ PHG2_VERSION }}"
+
+source:
+  url: https://github.com/maize-genetics/phg_v2/releases/download/{{ PHG2_VERSION }}/PHGv2-v2.2.tar
```
zrm22, line 8: So if we update the version to 2.3 does everything but this line update automatically?  Is there a way to get the PHGv2-v2.2.tar also automatically picked up during the release process?

matthewwiese: Actually I just did it now instead. Please see [this commit](https://github.com/maize-genetics/phg_v2/pull/67/commits/c6c94b8d2f2ccb889269cc2f08fad68c12319886).

btmonier (approved):

zrm22 (approved):

.github/workflows/build_upload_conda.yml
```diff
@@ -0,0 +1,33 @@
...
+  build-upload-conda:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Conda
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
+          python-version: 3.11
```
lynnjo, line 25: Is there a reason to use python 3.11 versus something newer?

.github/workflows/run_deploy_on_merge.yml
```diff
@@ -152,3 +152,23 @@ jobs:
             ${{ steps.matching_release.outputs.body }}
             ${{ env.COMMIT_MSG }}
 
+      # Update repository variables based on new release version and its MD5 hash
+      # These are used by the Conda package build Action
+      - uses: action-pack/set-variable@v1
+        with:
+          name: 'PHG2_VERSION'
+          value: '${{ env.VERSION }}'
+          token: ${{secrets.PHGV2CD}}
```
lynnjo, line 161: I realize "secrets.<>" is used in existing code.  Where are the secrets values stored (ie what file is accessed?). Do we have these created for maizegenetics.net, or is this per project?

lynnjo: Nevermind - I see from your slack posting this was generated for maize-genetics

.github/workflows/build_upload_conda.yml
matthewwiese: The newest version is 3.12 (3.13 is still prerelease) so we are pretty "bleeding edge" here. The Python version doesn't really matter all that much for this anyway, I simply included it so that it's explicit in case it breaks in the future.

Hugging Face [uses 3.8](https://github.com/huggingface/datasets/blob/main/.github/workflows/release-conda.yml#L28) in their v2 usage. If you go to the `setup-miniconda` repo their examples include a variety of versions.

.github/workflows/build_upload_conda.yml
```diff
@@ -0,0 +1,33 @@
...
+    types: [published]
+
+env:
+  ANACONDA_API_TOKEN: ${{ secrets.CONDA_TOKEN }}
+  PHG2_VERSION: ${{ vars.PHG2_VERSION }}
+  PHG2_RELEASE: ${{ vars.PHG2_RELEASE }}
+  PHG2_VERSION_MD5: ${{ vars.PHG2_VERSION_MD5 }}
+
+jobs:
+  build-upload-conda:
```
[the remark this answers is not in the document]
matthewwiese: Sure can do!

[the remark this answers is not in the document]
matthewwiese: [Let me know what you think](https://github.com/maize-genetics/phg_v2/pull/67/files?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

We can add more to the Quick Start in future - I wasn't sure how much we wanted included since stuff is still in flux.

.github/workflows/build_upload_conda.yml
lynnjo: My mistake/ I was thinking this was 2.11, not 3.11.  Ignore my comment/

.github/workflows/build_upload_conda.yml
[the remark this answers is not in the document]
lynnjo: Here is my concern:  If they pull the apps into the base conda environment, then the programs are seen via ProcessBuilder().  I think this is what you originally tested.  I did not understand if this worked when our Clikt commands precede the ProcessBuilder() commands with a conda environment setting.  Do all environments inherit what is in the base environment?  if yes, then this was ok.

But I think we agreed we don't want to encourage adding to the base environment.


If users create a new conda environment and load into that  this will not work with the ProcessBuilder() commands unless the environment is one that we know.  So either the user should always create a conda environment named phgv2-conda , or they will have to pass a new paramter with the conda environment name to every class that has a ProcessBuilder call.

let me know if I'm missing something here.

matthewwiese:
@lynnjo @zrm22 @aberthel @btmonier @pjbradbury @tcasstevens 

For everybody's benefit and to clear up any confusion, I want to respond to Lynn's question [from here](https://github.com/maize-genetics/phg_v2/pull/67#discussion_r1423049639) in a more visible way so that we are all on the same page.

If you look in the source, the ProcessBuilder commands follow this form:

```console
conda run -n phgv2-conda agc
```

The above runs the given program (in this case `agc`) *within* the Conda environment specified by the `-n` parameter, in this case the `phgv2-conda` environment that is created by `phg2 setup-environment`. This allows you to run any software installed in a given environment without having to be in the environment itself. Please refer to the `conda run` [docs here](https://docs.conda.io/projects/conda/en/latest/commands/run.html).

From a system with a brand new Conda installation, issue the following commands:

```console
conda create --name my_env --channel conda-forge --channel maize-genetics phg2
conda activate my_env
phg2 setup-environment
conda run -n phgv2-conda anchorwave --help
```

You should be greeted with the familiar AnchorWave help info. Furthermore, we can confirm the behavior I describe via:
```console
conda run -n phgv2-conda conda list
```

Which will run Conda within the `phgv2-conda` environment, listing all packages installed into said environment (all without having to be in the environment itself). The first line output should look something like the following depending on your OS:

```
# packages in environment at /home/matt/miniconda3/envs/phgv2-conda:
```

What I was trying to describe in Programmer Meeting is that if the dependencies are a part of the package installation itself, we wouldn't have to rely on a bespoke `phgv2-conda` environment. ProcessBuilder would find the software installed alongside PHG2, whether that be in the base environment or a user's own custom environment.

One might argue that users of the plain tarball application will no longer have access to the supplied dependencies, and to that I say: it's not a problem! For example, AnchorWave doesn't install minimap2 for you despite depending on it. We ought to provide a singular "blessed" installation method (Conda) while also providing a generic release (the tarballs) for advanced users. If somebody has little technical or bioinformatics knowledge, we direct them to installing the package via Conda, which will manage the dependencies for them and allow for a straightforward upgrade path. Advanced users can install the tarball and add its executable to their `PATH` if they wish - such users ought to be expected to have the requisite knowledge of downloading/building from source the other dependencies that we require.

Additionally, there are engineering concerns that I didn't have the time to explain at the meeting. Let's say a bug is discovered in one of our dependencies included in the `phgv2-conda` environment which requires an update to some latest version - existing users will have to manually delete and recreate the `phgv2-conda` environment, or enter it and update the problematic package themselves.

We should be using our tools/dependencies how they were designed to be used, doing it any other way will inevitably lead to bugs and breakage, which in a small lab is particularly harmful due to everybody's limits on time. Much like the unit and integration tests, thoughtful design and engineering will ultimately save us time and lead to better science.

.github/workflows/build_upload_conda.yml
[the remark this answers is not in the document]
matthewwiese: Please [see here](https://github.com/maize-genetics/phg_v2/pull/67#issuecomment-1850877235) for a response - I wanted to make sure everybody was on the same page and understood the discussion from this morning.

tcasstevens (approved):

README.md
```diff
@@ -7,7 +10,31 @@ The PHG is a trellis graph based representation of genic and intergenic regions
...
+
+```
+conda create --name my_env --channel conda-forge --channel maize-genetics phg2
+```
+
+If you've already installed PHG2, you can update to the newest version via:
+
+```
+conda update phg2
+```
```
lynnjo, line 33: Based on these changes, the documentation indicates the user can only execute phg from the conda setup (the original ./phg commands have been removed).  

Did we decide Conda is an option, or did we decide conda is the only option for installing phg?

If the latter, the instructions in the PHGv2- Building and Loading have an inconsistency.  Is our program named "phg" or "phg2" ? (we need to make a collective decision)

It looks like we'll have 2 environments - one that contains the phg(2) executable, and one that is created by phg(2) to run agc, tiledbvcf, etc.  Is that correct?

lynnjo: BTW, I'm not opposed to loading phg only via conda.  I just want to ensure our documentation has consistent examples and that we're all on the same page.

matthewwiese: > Did we decide Conda is an option, or did we decide conda is the only option for installing phg?

I don't think we decided anything, this is a question for @zrm22 or the group, same as the naming question.

> It looks like we'll have 2 environments - one that contains the phg(2) executable, and one that is created by phg(2) to run agc, tiledbvcf, etc. Is that correct?

There is only a single environment, `phgv2-conda`. The other environment containing the actual program (the Conda package) isn't really relevant as I wrote in my long comment above.

lynnjo (approved):

pjbradbury (approved):

README.md
zrm22: I think for now I would prefer explaining both the conda install and tarball installation instructions.  

I agree that long term if the user installs through conda we should only have one environment(no more need for SetupEnvironment) with all the dependencies, but I think it gets tricky as every time we make a call to anchorwave, tileDB, bgzip or bcftools we need to wrap the command that ProcessBuilder executes within a conda run command where we need to specify the environment name.  

The way we have this setup its consistent as we give the environment a name.  If the user needs to create their environment and then just add the PHG conda env to it, we no longer have this control.  

To fix either we need to have Processbuilder first figure out what envs are there and try to determine which one is the PHG one(what happens if you have more than 1?) or we have the user submit the name as a param(this would need to basically be in every command).  We may be able to have the user create an Environment variable that gets picked up by Clikt automatically, but it is hard to say.

README.md
```diff
@@ -1,4 +1,7 @@
 # PHG version 2
+> [!TIP]
```
zrm22, line 2: Put this tip under the badges.

README.md
```diff
@@ -27,56 +54,3 @@ The redesign leverages the powerful TileDB-VCF database, which is widely used in
     Composite Reference Haplotypes 
 
 More information on terminology can be found [here](docs/terminology.md).
-
-# Example usage
```
zrm22, line 31: These need to be brought back in.  We are making an effort so the user does not need to jump through 3-4 pages just to get basic information to run the software.  Obviously this is not all the documentation, but having the 10-15 commands that need to be run at the main page of the Repo is a good idea for what we are trying to do here.

zrm22 (changes requested):
We need to have the example usage section brought back into the README.  We are trying to show how simple it is to run through the pipeline and in my opinion the best way to do that is to show the full basic pipeline at the homepage of the repo.

PR #70, opened 2023-12-12
Chrom batching create fasta from hvcf

lynnjo:
## Description

## Type of change

## Checklist:

## Changelog entry

```release-note

```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
@@ -99,7 +99,8 @@ fun retrieveAgcData(dbPath:String,agcCmdList:List<String>):List<String>? {
     command.addAll(commandPrefix)
     command.addAll(genomes)
 
-    myLogger.info("retrieveAgcData: Running Agc Command:\n${command.joinToString(" ")}")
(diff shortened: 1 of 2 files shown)


tcasstevens (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
```diff
@@ -99,6 +99,7 @@ fun retrieveAgcData(dbPath:String,agcCmdList:List<String>):List<String>? {
     command.addAll(commandPrefix)
     command.addAll(genomes)
 
+    myLogger.info("retrieveAgcData: Running Agc Command:\n${command.joinToString(" ")}")
```
zrm22, line 102: I would remove this/shorten it.  If we have really long sets of regions this command will get very long and pollute the log.

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
lynnjo: I substringed this one, but it actually isn't the long one - this query at most has a list of samples from which we want to AGC to show us the contigs.

The long one is in queryAgc(), which is where we send the query with a whole bunch of ranges from which to pull sequence.  I added subsetting to that log message.

PR #87, opened 2024-01-18
Update maf vcf

aberthel:
## Description

Updated BioKotlin to v0.11. This changed gvcf generation, resulting in changes to hvcf generation that had to be incorporated.
A bug in hvcf generation was also discovered and fixed. 

## Type of change

## Checklist:

## Changelog entry

```release-note
Updated BioKotlin version to 0.11 and fixed hvcf generation that broke as a result of the update
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
@@ -14,6 +14,8 @@ import htsjdk.variant.vcf.VCFHeaderLine
 import htsjdk.variant.vcf.VCFHeaderVersion
 import net.maizegenetics.phgv2.utils.*
 import java.io.File
+import kotlin.math.max
+import kotlin.math.min
 
 data class HVCFRecordMetadata(val sampleName: String, val refSeq : String = "", val asmSeq : String = "",
                               val refContig : String, val refStart: Int, val refEnd: Int,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
```diff
@@ -288,43 +285,51 @@ class CreateMafVcf : CliktCommand(help = "Create g.vcf and h.vcf files from Anch
...
     }
 
     /**
-     * Function to see if the variant is partially contained at the start of the BED region
-     * Bed:      |--------------|
-     * Var:    |---|
+     * Function to see if the start of the variant is partially contained in the BED region
+     * Indels are left-justified
+     * Bed: |--------------|
+     * Var:              |---|
```
lynnjo, line 312: Functions variantPartiallyContainedStart() and variantPartiallyContainedEnd() were correct, but the diagrams in the comments above each were reversed?  What you've changed makes more sense based on the code that is checking the variant values against the region values.

aberthel: Yes. The functions were correct in behavior, but the comments describing the behavior were backwards (as I understood them) so I fixed it.

lynnjo (approved):

tcasstevens (approved):

btmonier (approved):

PR #119, opened 2024-03-18
Annotate fastas remove bgzip step

lynnjo:
## Description

## Type of change

## Checklist:

## Changelog entry

```release-note
Annotate fasta no longer compresses the new file as anchorwave needs uncompressed fastas
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/agc/AnnotateFastas.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/agc/AnnotateFastas.kt
@@ -143,31 +143,8 @@ class AnnotateFastas : CliktCommand(help = "Annotate FASTA files with sample nam
                             writer.write("${line}\n")
                         }
                     }
-//                    if (entry.compressed) {
-//                        // If the input file was compressed, then the new output file will be compressed
-//                        gzipFasta(newFilename)
-//                    }
                 }
             }
         }
-    fun gzipFasta (file:String) {
-        try {
 
-            // gzip the file
-            // use the -f option to overwrite any existing file
-            myLogger.info("gzipping  file ${file}")
-            var builder = ProcessBuilder("conda","run","-n","phgv2-conda",
-                "gzip", "-f", file)
-
-            var process = builder.start()
-            var error: Int = process.waitFor()
-            if (error != 0) {
-                myLogger.warn("\nERROR $error creating gzipped  version of file: $file")
-                throw IllegalStateException("bgzipAndIndexGVCFfile: error trying to gzip file ${file}: ${error}")
-            }
-
-        } catch (exc:Exception) {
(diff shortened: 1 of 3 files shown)


zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/agc/AnnotateFastas.kt
```diff
@@ -125,14 +126,10 @@ class AnnotateFastas : CliktCommand(help = "Annotate FASTA files with sample nam
...
 
-                // if file was previously compressed, we compress the new file.
-                // Strip off the .gz extension before writing the new non-compressed file.
-                // it wil be added when we re-compress the file.
-                // The file names will either be <sampleName>.fa or <sampleName>.gz
-                // When the file is compressed, we only add .gz as an extension as AGC
-                // takes everything before the last period as the sample name and we
-                // want the sample name to look like "B73" or "CML247" and not "B73.fa" or "CML247.fa"
-                val justName = if (fastaFile.endsWith(".gz")) sampleName else "${sampleName}.fa"
+                // Previously if the original file was previously compressed, we compressed the new file.
```
tcasstevens, line 129: Should this say wasn't?

src/main/kotlin/net/maizegenetics/phgv2/agc/AnnotateFastas.kt
```diff
@@ -125,14 +126,10 @@ class AnnotateFastas : CliktCommand(help = "Annotate FASTA files with sample nam
...
+                // We no longer do this, as the align-assemblies step requires uncompressed fasta files.
+                // The file names will  be <sampleName>.fa
+                val justName =  "${sampleName}.fa"
```
tcasstevens, line 132: What if the incoming file is already compress?

tcasstevens (approved):

src/main/kotlin/net/maizegenetics/phgv2/agc/AnnotateFastas.kt
lynnjo: No, it is correct other than I should remove one of the "previously".

In the original code:  if the original fasta file was compressed, I compressed the annotated fasta.  If the original was uncompressed, I did not compress the created file.  

Now I am never compressing the annotated files as these files are input for align-assemblies and anchorwave does not handle compressed files.

src/main/kotlin/net/maizegenetics/phgv2/agc/AnnotateFastas.kt
lynnjo: This code will never compress the output fastas as they are used in the align-assemblies step, and anchorwave does not handle compressed files.  I didn't think of that when I originally wrote the code - I was only thinking of AGC which can handle the compressed files.

pjbradbury (approved):

PR #121, opened 2024-03-18
Agc compress clarify parameters

lynnjo:
## Description

## Type of change

## Checklist:

## Changelog entry

```release-note
Update command help messages to clarify when annotated fastas are needed.
```


src/main/kotlin/net/maizegenetics/phgv2/cli/AgcCompress.kt
```diff
@@ -37,23 +43,23 @@ import java.time.LocalDate
...
+class AgcCompress : CliktCommand(help = "Create a single AGC compressed file from an input of FASTA files created with the phg annotate-fasta command") {
 
     private val myLogger = LogManager.getLogger(AgcCompress::class.java)
 
     val dbPath by option(help = "Folder name where AGC compressed files will be created.  Should be the same parent folder where tiledb datasets are stored." )
         .default("")
 
 
-    val fastaList by option(help = "File containing full path name for the fasta files, one per line, to compress into a single agc file.  Fastas may be compressed or uncompressed files. Reference fasta should NOT be included.")
+    val fastaList by option(help = "File containing full path name for the annotated fasta files, one per line, to compress into a single agc file.  Fastas may be compressed or uncompressed files. Reference fasta should NOT be included.\nAll fastas (including reference) must be fastas created via the phg annotate-fastas command")
```
tcasstevens, line 54: This says "Reference fasta should NOT be included". But next sentence says "including reference"?

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
```diff
@@ -290,6 +290,7 @@ fun queryAgc(commands:Array<String>):Map<Pair<String,String>,NucSeq> {
                         // samplename/genome in the VCF files.
                         myLogger.error("queryAgc: Error:  No sampleName= found in idline: returned from AGC query.")
                         myLogger.error("Please run the AnnotateFastas command to add sampleName to your fasta files.")
+                        myLogger.error("Then delete the assemblies.agc file from you tiledb folder and recreate the agc compressed file via the agc-compress command using the new annotated fastas.")
```
tcasstevens, line 293: Should we consider deleting that file automatically?

tcasstevens (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
lynnjo: I think that would be an excellent idea, except there is apparently no way to remove a fasta from AGC.  Seems the way they have stored the compressed files, everything is intertwined.  In fact, when you append a file, it doesn't really append, but instead drops the old compressed file and re-creates it with the new fastas included.

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
lynnjo: hmm ... what are the down sides to this?  It seems they will have to recreate the file, so us deleting assemblies.agc is fine.  There are backups to consider:

1.  user has just created the agc file and didn't use the annotated fastas.  Deleting the assemblies.agc makes sense.  User starts fresh.
2. user previously created the agc file with the appropriate annotated fastas.  Then they did an "append" - adding a new assembly and they forgot to use the annotated version.  It should still be ok to delete the assemblies.agc file as we (phg code) writes a backup before each append list (because agc deleltes/recreates the compressed file).  If it were an informed user (like me) I would know I could copy the most recent backup to assemblies.agc and re-run my latest command using the annotated-fasta.

So the code could do the backup copy.  This gets tricky if they did multiple iterations of appending before running any analysis, and don't know which times they used annotated fastas and which times they didn't.

Code could delete and restore a backup, back we don't know which one if any backups are good, so probably shouldn' do this.

tcasstevens: I'm not sure what should happen, but the error message says "delete the assemblies.agc".  Does the user know there may be reasons not to do that?

lynnjo: If we tell them to delete the assemblies.agc, we also need to explain they have to re-run agc-compress for all of their fasta files.

if we don't (or they don't) delete the assemblies.agc file, the AGC code will try to append to the existing file.  It will fail when re-adding the same sample name because that already exists (because it takes the samplename from the file name).  Or it might not fail if the file name is something long that annotate- fastas renamed to something short.

I feel I need to write a book of instructions for this scenario!

AGC compress is quick - perhaps we just delete the file and tell them they have to re-run agc-compress for all their samples, making sure they use annotated files.  Would this make sense to you?

pjbradbury (approved):

src/main/kotlin/net/maizegenetics/phgv2/utils/SeqUtils.kt
pjbradbury: It seems like this should be checked before adding fasta to assemblies.agc. For existing assemblies.agc that were built incorrectly, I think telling the user to delete is okay. That's the only way in a non-interactive command-line version to get the users permission to delete the file.

lynnjo: We can check all the fastas but will be a performance hit.  Currently, the code takes the list of fastas and gives them to AGC - agc does the rest and phg does nothing but wait for agc to finish.

if the user has 100 or more fastas they want to use for creating the agc compressed file, this means opening up each one of them and checking the id lines.  THough we could probably check for first occurrence and if any are found, assume all idlines are good.  I guess it only happens once each time we load so the hit won't be bad.

i'll do a few tests with the 29 fastas we load to our test instance and see if there are any issues.

tcasstevens: Can something check this without changing the assemblies.agc until we're fairly certain it will succeed?

btmonier (approved):

tcasstevens (approved):

PR #138, opened 2024-04-04
Anchorwave dotplot

lynnjo:
## Description

Add creation of a dot plot summary of anchorwave alignments.  This is a basic summary, specific to data from the anchorwave generated anchorspro file, created for each alignment.

## Type of change

## Checklist:

## Changelog entry

```release-note
Add dot plot png creation, per assembly, when anchorwave is run.
```


build.gradle.kts
```diff
@@ -39,6 +39,12 @@ dependencies {
...
     val kotlinVersion = rootProject.extra["kotlinVersion"]
 
+    implementation("org.jetbrains.kotlinx:dataframe:0.13.1")
+    implementation("org.jetbrains.lets-plot:lets-plot-kotlin:4.7.0")
+
+    implementation("org.jetbrains.lets-plot:lets-plot-kotlin-jvm:4.7.0")
+    implementation("org.jetbrains.lets-plot:lets-plot-image-export:4.3.0")
+    implementation("org.jetbrains.lets-plot:lets-plot-batik:4.3.0")
     implementation("org.biokotlin:biokotlin:0.14")
     implementation("com.github.ajalt.clikt:clikt:4.2.0")
```
tcasstevens, line 50: See below implementation("org.jetbrains.kotlin:kotlin-test:1.9.10")
Maybe that should use ${kotlinVersion}
I know I did it, but not sure why

tcasstevens (approved):

src/main/kotlin/net/maizegenetics/phgv2/cli/AlignAssemblies.kt
```diff
@@ -581,6 +598,60 @@ class AlignAssemblies : CliktCommand(help = "Align prepared assembly fasta files
...
+     * showing basic alignment data in scatter plot form with the x-axis being the query
+     * sequence start position, the y-axis being the reference sequence start position, and
+     * the color of the data points based on the strand (red=forward, blue=reverse).
+     * The plot is faceted by the query and reference chromosomes.
+     *
+     * For  plots based on specific reference or query ids, users can use the plotDot function
+     * from rPHG.
+     *
+     */
+    fun plotDot(
```
btmonier, line 635: I wonder if, in the future, it would be better to place these visualization methods in some sort of utility class?

btmonier (approved):

src/main/kotlin/net/maizegenetics/phgv2/cli/AlignAssemblies.kt
lynnjo: @btmonier I initially had this in a Metrics class, but moved it back to AlignAssemblies.  This particular visualization is very specific to the anchorwave generated anchorspro file so I think it belongs here.  We could create an altered version of this in the future that has a more generic appeal, and that could go into a metric file.

build.gradle.kts
lynnjo: I changed to use kotlinVersion

aberthel (approved):

PR #157, opened 2024-05-01
Qc documentation

aberthel:
## Description

## Type of change

- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Checklist:

## Changelog entry

```
added documentation for anchorwave and vcf-based QC metrics
```


docs/qc_metrics.md
```diff
@@ -0,0 +1,135 @@
...
+
+1. As part of `create-maf-vcf`:
+
+```shell
+phg create-maf-vcf \
+    --db-path vcf_dbs \
+    --bed output/ref_ranges.bed \
+    --reference-file output/updated_assemblies/Ref.fa \
+    --maf-dir output/alignment_files \
+    -o output/vcf_files \
```
lynnjo, line 69: It always seems confusing that most of the parameters are prefixed with 2 dashes, but output is with 1.  Perhaps that is just to accept.

docs/build_and_load.md
```diff
@@ -975,7 +975,9 @@ phg create-maf-vcf \
     --bed output/ref_ranges.bed \
     --reference-file output/updated_assemblies/Ref.fa \
     --maf-dir output/alignment_files \
-    -o output/vcf_files
+    -o output/vcf_files \
+    --metrics-file output/vcf_files/VCFMetrics.tsv \
```
tcasstevens, line 979: Seems like the absence of a metrics file would imply --skip-metrics

tcasstevens (approved):

docs/build_and_load.md
lynnjo: @tcasstevens I think the metrics-file parameter is used to change the name and location of where this file will be printed.  Its absence means to us default location, but it will still be created unless skip-metrics is true.

lynnjo (approved):

docs/qc_metrics.md
tcasstevens: @lynnjo That is the convention for command line parameters. "long for of a flag" has 2 dashes. Short form (i.e. one letter) has 1 dash

docs/build_and_load.md
aberthel: @tcasstevens @lynnjo is correct. It's set up this way because we wanted the default behavior to produce a metrics file, to encourage users to do QC as they go.

pjbradbury (approved):

PR #182, opened 2024-07-19
Multiple file bug create fasta from hcvcf

lynnjo:
## Description

Handle creating fasta from hvcf when input is a directory of hvcf files

## Type of change

## Checklist:

## Changelog entry

```release-note
Fixed bug where createFastaFromHvcf processed a directory of files and only wrote to 1 fasta.
```


src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -40,11 +43,11 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a FASTA file from a h.v
     val dbPath by option(help = "Folder name where TileDB datasets and AGC record is stored.  If not provided, the current working directory is used")
         .default("")
 
-    val output by option("-o", "--output", help = "Name for output Fasta file")
+    val outputDir by option("-o", "--output-dir", help = "DIrectory where output fasta FIles will be written")
```
zrm22, line 46: Typos in Directory and in Files

zrm22: Also does this need both names?  or will --output-dir be automatically picked up?  A unit test could check this quickly.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -66,33 +69,55 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a FASTA file from a h.v
...
-                writeSequences(output,hvcfFiles.flatMap { processSingleHVCF(VCFFileReader(it,false), dbPath,condaEnvPrefix) }
-                    .associate { it.id to it }
-                    .values.toList(), fastaType)
+    fun buildFastaFromHVCF(dbPath: String, outputDir: String, fastaType:String, hvcfDir: String ,hvcfFile : String, condaEnvPrefix:String) {
+
+        if (hvcfDir != "") {
+            // Loop through the directory and figure out which files are hvcf files
+            // The gvcf and hvcf files may be in the same folder, so verify specific extension
+            val hvcfFiles = File(hvcfDir).listFiles { file ->
+                file.name.endsWith("h.vcf") || file.name.endsWith("h.vcf.gz") || file.name.endsWith("hvcf") ||
```
zrm22, line 85: Should you use the REGEX here?

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -66,33 +69,55 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a FASTA file from a h.v
...
+                        file.name.endsWith("hvcf.gz")
+            }
+
+            // Loop through each file and run the processSingleHVCF function
+            hvcfFiles?.forEach { hvcfFile ->
+                val tempFileName = hvcfFile.name.replace(HVCF_PATTERN, ".fa")
```
zrm22, line 91: Can we get rid of tempFileName?  From a quick glance its only used to build the outputFileName.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -66,33 +69,55 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a FASTA file from a h.v
...
+                val outputFileName = File(tempFileName).name.replace(".fa", "_${fastaType}.fa")
+                val outputFile = "$outputDir/$outputFileName"
+
+                BufferedWriter(FileWriter(outputFile)).use { output ->
+                    writeSequences(output,
+                        processSingleHVCF(VCFFileReader(hvcfFile, false), dbPath, condaEnvPrefix)
+                            .associate { it.id to it }
```
zrm22, line 98: Is moving the data to a MAP needed?  Can you just go directly from processSingleHVCF -> List?  Or do you expect to have multiple records in the output that have the same id?

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -66,33 +69,55 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a FASTA file from a h.v
...
+            // Load in the TileDB
             TODO("TileDB VCF Reader Not implemented yet.  Please run with --hvcf-file")
         } else {
-            //Load in the HVCF
-            val hvcfFile = VCFFileReader(File(hvcfFile),false)
+            // Load in the HVCF
+            val hvcfFileReader = VCFFileReader(File(hvcfFile), false)
+            val tempFileName = File(hvcfFile).name.replace(HVCF_PATTERN, ".fa")
+            val outputFileName = File(tempFileName).name.replace(".fa", "_${fastaType}.fa")
+            //val outputFileName = "${fileHvcf.nameWithoutExtension}.fa"
```
zrm22, line 112: Remove comment.  Also similar to above can we get rid of tempFileName

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateRefVcf.kt
```diff
@@ -190,14 +187,12 @@ class CreateRefVcf : CliktCommand(help = "Create and load to tiledb a haplotype
...
                     if (!currentSeqHashs.contains(intervalHash)) {
-                        altHeaderLines.add(
-                            VCFAltHeaderLine(
-                                "<ID=${intervalHash}, Description=\"haplotype data for line: ${refName}\"," +
-                                        "Source=\"${refGenome}\",SampleName=\"${refName}\",Regions=\"${chr}:${anchorStart}-${anchorEnd}\"," +
-                                        "Checksum=\"${intervalHash}\",RefChecksum=\"${intervalHash}\",RefRange=\"${chr}:${anchorStart}-${anchorEnd}\">",
-                                VCFHeaderVersion.VCF4_2
-                            )
-                        )
+                        val altHeaderLine = StringBuilder().apply {
```
zrm22, line 190: Is there a reason to change this to a StringBuilder?  The template that it was doing before should be fine.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
lynnjo: This command was originally written with both options, which yes, I think should be mutually exclusive.  I ignored that while fixing the output bug, but could address it as part of this pull request.

src/test/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcfTest.kt
```diff
@@ -349,28 +350,110 @@ class CreateFastaFromHvcfTest {
...
-        val outputFastaHaplotypes = NucSeqIO("${TestExtension.testOutputFastaDir}/Ref_Test_output_hap.fa").readAll()
+        val outputFastaHaplotypes = NucSeqIO("${TestExtension.testOutputFastaDir}/Ref_haplotype.fa").readAll()
+        outputFastaHaplotypes.values.map { getChecksumForString(it.seq()) }.toSet().forEach{
+            assertTrue(truthHashes.contains(it))
+        }
+
+    }
+
+    @Test
+    fun testBuildFastaFromHVCF_dirInput() {
```
zrm22, line 376: You should use kotlin naming conventions here with camelCase not _.  Same for the single file one as well.

src/test/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcfTest.kt
```diff
@@ -349,28 +350,110 @@ class CreateFastaFromHvcfTest {
...
+        // way the Ref.h.vcf file was created.  The haplotype sequence in the LineB.h.vcf file should be the same
+        // sequence as the original LineB.fa file.   This file has been created and stored to data/test/createFastaFromHvcf/LineB.h.vcf
+
+        val refHVCFFileName = "data/test/smallseq/Ref.h.vcf"
+        val lineBHvcfFileName = "data/test/createFastaFromHvcf/LineB.h.vcf"
+
+        // Write the Ref and LineB hvcf files to the output directory: TestExtension.testOutputHVCFDir
+        val hvcfDir = TestExtension.testOutputHVCFDir
+        File(hvcfDir).mkdirs()
+        File(TestExtension.testOutputHVCFDir+"/Ref.h.vcf").writeBytes(File(refHVCFFileName).readBytes())
```
zrm22, line 402: Consider using Path.copyTo

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateRefVcf.kt
lynnjo: Yes, the StringBuilder() is more efficient as it is not creating a new String object with each concatenation.  This code was changed based on some junit code (for creating the LIneB.h.vcf file) that ended up not being needed.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
lynnjo: yes, I should have replaced that code when I created the REGEX,

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateRefVcf.kt
zrm22: The only reason for the concatenation is just for readability in the IDE.  My hunch is the kotlin compiler will merge this together.  I feel that it is much more readable without the StringBuilder as the String with templating is much simpler.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
lynnjo: yes, updated

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
lynnjo: updated

src/test/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcfTest.kt
lynnjo: The name is camel-cased except at the end.  I used the underscore to make it clear that one case is testing input from  a directory of files and the other is just a single file.  I prefer to keep the names as they are currently as it is much easier to identify the 2 different test cases.

src/test/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcfTest.kt
lynnjo: updated

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
lynnjo: I misread your comment .  You were asking if both -o and --output-dir were needed.  Don't know, will test.  My change was to match outputDir with output-dir, which replaced the previous "output" parameter.

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
lynnjo: no, map isn't needed.  All that is needed for that parameter is the output from processSingleHVCF().  Code updated.

tcasstevens (approved):

PR #203, opened 2024-08-02
Update maf to gvcf

zrm22:
## Description

Update MAFToGVCF code to call the new version of biokotlin.  The MAF file created by anchorwave 1.2.3 is different and this allows the user to input either older or current versions of MAF files.

## Type of change

## Checklist:

## Changelog entry

```release-note
Update Maf to GVCF conversion by using biokotlin 0.19.
```


README.md
```diff
@@ -47,6 +47,11 @@ _Long-form documentation for this section can be found
 [here](docs/build_and_load.md). Additional information about **QC 
 metrics** can be found [here](docs/qc_metrics.md)._
 
+> [!NOTE]
+> As of version 2.4.X, the phg utilizes a new version of anchorwave(1.2.3).
+> This changes how ASM coordinates are handled.  If you are using old MAF files,
```
lynnjo, line 52: I would state this more explicitly.  "If you are using MAF files generated with an anchorwave version earlier than 1.2.3, or from a phgv2 version earlier than ...." please use the --legacy-maf-file ....

docs/build_and_load.md
```diff
@@ -1,5 +1,11 @@
 # Building and Loading
 
+!!! note
+    As of version 2.4.X, the phg utilizes a new version of anchorwave(1.2.3).
+    This changes how ASM coordinates are handled.  If you are using old MAF files,
```
lynnjo, line 5: I would change to "If you are using MAF files created with a phgv2 version earlier than 2.4.0 ..."

docs/build_and_load.md
```diff
@@ -1198,6 +1204,10 @@ In addition, `--conda-env-prefix` is an optional parameter that specifies the pa
 to the Conda directory that contains the conda environment needed to run phg.
 If not set, conda env `phgv2-conda` in the defauilt location will be used.
 
+There is also an optional parameter: `--legacy-maf-file`.  This flag can be added to the command
+to be able to process MAF files created from anchorwave 1.2.2 or earlier.  If this is not set and
```
lynnjo, line 1208: I've been thinking users won't know they phg version, but they will if they know what is in their phg_environment.yml file.  Or if they ran "anchorwave --version".  

Should we tell them explicitly how to check their anchorwave version?  Let them know that merely downloading the new PHG does not update their conda environment, but that they need to run commands to explicitly update it?

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateMafVcf.kt
```diff
@@ -4,10 +4,7 @@ import biokotlin.genome.*
 import biokotlin.seq.NucSeq
```
tcasstevens, line 4: Change the parameters to use required()?

tcasstevens (approved):

README.md
```diff
@@ -47,6 +47,12 @@ _Long-form documentation for this section can be found
 [here](docs/build_and_load.md). Additional information about **QC 
 metrics** can be found [here](docs/qc_metrics.md)._
 
+> [!NOTE]
+> As of version 2.4.X, the phg utilizes a new version of anchorwave(1.2.3).
+> This changes how ASM coordinates are handled. 
+> If you are using old MAF files generated either from anchorwave 1.2.2 or from PHGv2 version 2.3 or eariler, 
+> please use the --legacy-maf-file flag for create-maf-vcf.  More information can be found [here](docs/build_and_load.md).
+
```
lynnjo, line 55: I like the new wording.  

But I'm still concerned users may not grasp that running with the new phg version does not mean that any new maf files created from align-assemblies will be with the newer version of anchorwave.  We need to make it clear to them somewhere in the documentation that they must update their conda environment with the new version of anchorwave as is specified in the phg_environment.yml file.  Just re-running setup-environment won't fix this either as it skips the setup if the "phgv2-conda" env already exists.   They will need to do a "conda update ..."

@btmonier Do you have thoughts on where best this information should be documented?

lynnjo (approved):

README.md
zrm22: I can add in a note that it is recommended to remake your conda environment, but users should really be doing this every time they are downloading a new version anyways to fully rule out inconsistencies.  I am fully expecting we get some user requests asking why the coordinates are different, but unfortunately there really is not a good way to force people to do it right.  I am also going to put out an announcement on the discussions page once this branch is merged in.

AGanesan12 (approved):

README.md
btmonier: @lynnjo @zrm22 - if anything, we could also add a note in the [Setup Environment ](https://phg.maizegenetics.net/build_and_load/#set-up-conda-environment) section of the docs.

btmonier (approved):

README.md
lynnjo: @btmonier I think adding a note in the SetupEnvironment section is a good idea.

docs/build_and_load.md
aberthel: I think "how to check the version of software I have installed" is general enough information that we don't need to include it. Especially since we're using conda and any search to that effect returns `conda list` as a viable option. But I agree that reminding them they need to update the conda environment manually can't hurt.

aberthel (approved):

PR #206, opened 2024-08-07
Handling gzipped file names correctly now.  Also adding in a helper f…

zrm22:
…unction to build kmers for later use.

## Description

Simple change so that the read mapping file has the correct name when the fastqs are gzipped.

## Type of change

## Checklist:

## Changelog entry

```release-note
Simple change so that the read mapping file has the correct name when the fastqs are gzipped.
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/pathing/AlignmentUtils.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/pathing/AlignmentUtils.kt
@@ -65,8 +65,14 @@ class AlignmentUtils {
                 //export the read mapping to disk
                 //Use the first file name as the readmapping output name
                 val inputFile1 = File(keyFileRecord.file1)
-                //Todo does this need to work with .fastq.gz files?
-                val outputFileName = "${outputDir}/${inputFile1.nameWithoutExtension}_readMapping.txt"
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/maizegenetics/phgv2/pathing/AlignmentUtils.kt
```diff
@@ -454,6 +460,29 @@ class AlignmentUtils {
             }
         }
 
+        /**
+         * Function that builds a set of minHashes from a sequence.
+         * This will work by doing a single pass over the read sequence and building the kmer hash set.
+         */
+        fun extractMinHash(sequence: String) : Set<Long> {
```
tcasstevens, line 467: Is this being called somewhere?

zrm22: Not yet.  I am using it for generating my training dataset for my Kmer Model but it was sitting in the same file as the other change so it was easier to just commit.  If it does not pan out I will remove it later.

tcasstevens (approved):

lynnjo (approved):

btmonier (approved):

PR #212, opened 2024-08-21
Create fasta from hvcf fix duplicate entries

lynnjo:
## Description

 When creating a haplotype fasta from an hvcf file, there should not be duplicates.  Fixed the code so that resulting fasta files will contain no more than 1 instance of a haplotype with its sequence.

## Type of change

## Checklist:

## Changelog entry

```release-note
CreateFastaFromHvcf now correctly contains a single instance of a haplotype sequence when when creating haplotype fastas.
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
@@ -140,7 +140,7 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a FASTA file from a h.v
         if(fastaType == FastaType.composite)
             writeCompositeSequence(outputWriter, haplotypeSequences)
         else if(fastaType == FastaType.haplotype) {
-            writeHaplotypeSequence(outputWriter, haplotypeSequences)
+            writeHaplotypeSequence(outputWriter, haplotypeSequences.toSet())
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
```diff
@@ -263,7 +263,8 @@ class CreateFastaFromHvcf : CliktCommand( help = "Create a FASTA file from a h.v
      * This is almost identical to how fastas were exported in the original version of the pipeline.
      */
     fun writeHaplotypeSequence(outputFileWriter: BufferedWriter, haplotypeSequences: List<HaplotypeSequence>, exportFullIdLine : Boolean = true) {
-        for(hapSeq in haplotypeSequences) {
+        // Make this a set to remove duplicates
+        for(hapSeq in haplotypeSequences.toSet()) {
```
zrm22, line 267: Should this be sorted?  Just so the output file is consistent given an input?

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/cli/CreateFastaFromHvcf.kt
lynnjo: hapSeq is the data class HaplotypeSequence.  If you sort by ID, you are sorting by values of the sequence hash.  That won't get you the data in the order of chrom/pos, so I'm not sure this is useful.

lynnjo: I could pehaps change to a LinkedHashSet or something that does maintain order.  I will test with that.

tcasstevens (approved):

aberthel (approved):

PR #221, opened 2024-09-12
Add showDefaultValues=true to clikt help messages

tcasstevens:
## Description

Add showDefaultValues=true to clikt help messages

## Type of change

- [X] `ENHANCEMENT` (non-breaking change which improves existing functionality)

## Checklist:

## Changelog entry

```release-note
Add showDefaultValues=true to clikt help messages
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/cli/Phg.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/cli/Phg.kt
@@ -17,7 +17,7 @@ class Phg : CliktCommand() {
         setupDebugLogging()
 
         context {
-            helpFormatter = { MordantHelpFormatter(it, showRequiredTag = true) }
+            helpFormatter = { MordantHelpFormatter(it, showRequiredTag = true, showDefaultValues = true) }
         }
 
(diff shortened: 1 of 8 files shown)


src/main/kotlin/net/maizegenetics/phgv2/cli/Phg.kt
```diff
@@ -17,7 +17,7 @@ class Phg : CliktCommand() {
         setupDebugLogging()
 
         context {
-            helpFormatter = { MordantHelpFormatter(it, showRequiredTag = true) }
+            helpFormatter = { MordantHelpFormatter(it, showRequiredTag = true, showDefaultValues = true) }
```
zrm22, line 20: This is a good idea.  What happens in things like map-kmers where some of the parameters already have Default is X messages in the help text?  You should probably go through and fix everything as part of this so its consistent.

tcasstevens: Ok, I removed the hardcoded default values

zrm22 (approved):

PR #229, opened 2024-10-02
Kmer mapping diagnostic - Initial PR

zrm22:
## Description

This is an initial set of updates to the read mapping section.  We are still working on building the integration test and fixing the imputation accuracies, but this contains some useful updates which simplify the current algorithm and give very similar results.  It also removes excessive logging from the building index and read mapping portions.

## Type of change

## Checklist:

## Changelog entry

```release-note
Updating imputation to remove excessive logging and to handle paired reads together instead of separate.
```


src/test/kotlin/net/maizegenetics/phgv2/pathing/MapKmersTest.kt
```diff
@@ -745,6 +749,20 @@ class MapKmersTest {
 
     }
 
+    @Ignore
+    @Test
+    fun testExtractKmersForPairedReads() {
```
zrm22, line 754: The chunks of code these 2 tests will eventually cover are covered by other tests so I am ignoring for now.  Will update these once the integration test is running smoothly.

src/main/kotlin/net/maizegenetics/phgv2/api/HaplotypeGraph.kt
```diff
@@ -114,6 +114,14 @@ class HaplotypeGraph(hvcfFiles: List<String>) {
         return result
     }
 
+    /**
+     * Simple function to make a map of all the haplotypeIds and all the SampleGametes which have that haplotypeId.
```
lynnjo, line 118: Is this included as a  more concise version of the original hapIdsToSamleGametes() function that previously existing in HaplotypeGraph?

zrm22: Yes, it allows users to just get the whole map if need be without having to loop through the refRanges.

smm477 (approved):

aberthel (approved):

lynnjo (approved):

PR #242, opened 2024-10-30
added ref and chosen haplotype len to ImputationMetrics

lynnjo:
## Description

Added columns RefLen and HaplotypeLen to the ImputationMetrics output

## Type of change

## Checklist:

## Changelog entry

```release-note
Added RefLen and HaplotypeLen columns to ImputationMetrics output
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/pathing/ImputationMetrics.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/pathing/ImputationMetrics.kt
@@ -165,7 +165,7 @@ class ImputationMetrics : CliktCommand(help = "Impute best path(s) using read ma
                 val refRange = bedRanges[pos]
                 val hapid = wgsPosToHapid.getOrDefault(pos, "NA")
                 val inImpuptation = if (hapid == "NA") "No" else "Yes"
-                val inP39 = if (parentChromPosToHapid.containsKey(pos)) "Yes" else "No"
+                val inParent = if (parentChromPosToHapid.containsKey(pos)) "Yes" else "No"
                 // hapid always has a value: NA if not present, and NA won't be in hapidToSampleName, so this works
                 val sampleName = hapidToSampleName.getOrDefault(hapid, "NA")
                 // get readCount for the hapid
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/maizegenetics/phgv2/pathing/ImputationMetrics.kt
```diff
@@ -164,8 +175,10 @@ class ImputationMetrics : CliktCommand(help = "Impute best path(s) using read ma
                 val parentSampleRdCt = if (parentSampleHapid == "NA") 0 else hapidCounts.getOrDefault(parentSampleHapid, 0)
                 // check if the hapid and parentSampleHapid are the same
                 val hapidMatch = if (hapid != "NA" && parentSampleHapid != "NA" && hapid == parentSampleHapid) "true" else "false"
+                val refLen = hapidToSeqSize.getOrDefault(hapid, Pair(0,0)).second
+                val hapLen = hapidToSeqSize.getOrDefault(hapid, Pair(0,0)).first
 
-                writer.write("$refRange\t$hapid\t$sampleName\t$inImpuptation\t$inP39\t$parentSampleHapid\t$sampleRdCount\t$parentSampleRdCt\t$hapidMatch\n")
+                writer.write("$refRange\t$refLen\t$hapLen\t$hapid\t$sampleName\t$inImpuptation\t$inP39\t$parentSampleHapid\t$sampleRdCount\t$parentSampleRdCt\t$hapidMatch\n")
```
zrm22, line 181: Shouldn't inP39 be something like in target or something like that?

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ImputationMetrics.kt
lynnjo: I thought I had fixed that - let me see where it got messed up.

lynnjo: Looks like I copied a wrong line - I"ll fix it.

lynnjo: Acutally, this is correct.  If you look at the code above, the variable name is "inP39", which is just a true/false value.  I will name the variable something different so it isn't confusing.

tcasstevens (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ImputationMetrics.kt
lynnjo: Note the actual column is defined as "In_${sampleHvcfName}", so in my test case it will show as "In_P39" (because P39 is my supposed parent), but this will vary based on user's sample name.

aberthel (approved):

PR #313, opened 2025-04-30
In Hvcf2Gvcf, parsing filename to get sample name using ".h.vcf" inst…

tcasstevens:
…ead of "."

## Description

## Type of change

- [X] `BUGFIX` (non-breaking change which fixes an issue)

## Checklist:

## Changelog entry

```release-note
In Hvcf2Gvcf, parsing filename to get sample name using ".h.vcf" instead of "."
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/cli/Hvcf2Gvcf.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/cli/Hvcf2Gvcf.kt
@@ -105,7 +105,7 @@ class Hvcf2Gvcf :
             .forEach { hvcfFile ->
                 myLogger.info("buildGvcfFromHvcf: Processing hvcf file: ${hvcfFile.name}")
                 var time = System.nanoTime()
-                val sample = hvcfFile.toString().substringAfterLast("/").substringBefore(".h.vcf")
+                val sample = hvcfFile.name.substringBefore(".h.vcf")
                 val records =
                     processHVCFtoVariantContext(sample, refSeq, outputDir, hvcfFile, dbPath, condaEnvPrefix, batchSize)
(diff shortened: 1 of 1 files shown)


src/main/kotlin/net/maizegenetics/phgv2/cli/Hvcf2Gvcf.kt
```diff
@@ -105,7 +105,7 @@ class Hvcf2Gvcf :
             .forEach { hvcfFile ->
                 myLogger.info("buildGvcfFromHvcf: Processing hvcf file: ${hvcfFile.name}")
                 var time = System.nanoTime()
-                val sample = hvcfFile.toString().substringAfterLast("/").substringBefore(".")
+                val sample = hvcfFile.name.substringBefore(".h.vcf")
```
zrm22, line 108: I am wondering if we need to check for extensions like the following:
`.hvcf`
I believe we support this extension in most if not all other places as an alternative for .h.vcf

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/cli/Hvcf2Gvcf.kt
tcasstevens: In the code above that, we're only checking these. Do you want me to add .hvcf?
.filter { it.name.endsWith("h.vcf.gz") || it.name.endsWith("h.vcf") }

zrm22: Yeah This will work with the files made from the PHG, but if it's anything like gvcfs, the extension is usually either .gvcf or .g.vcf, so we should probably handle both

PR #363, opened 2026-04-28
Read counts by sample

pjbradbury:
## Description

Adds the option to report read mapping counts by sample (haploid genome).

## Type of change

## Checklist:

## Changelog entry

Adds an option to report read mapping counts by reference range and sample name (haploid genome name).

```release-note

```


src/main/kotlin/net/maizegenetics/phgv2/pathing/ReadMappingCountQc.kt
```diff
@@ -20,18 +21,22 @@ class ReadMappingCountQc : CliktCommand("Read mapping count QC") {
...
         .required()
 
-    val targetSampleName by option(help = "Target sample name")
+    val targetSampleName by option(help = "Target sample name. Required but not used when the --by-sample flag is set.")
         .required()
 
     val outputDir by option(help = "Output directory")
         .required()
 
+    val bySample by option(help = "Create a report with a read count for each sample by reference range. " +
```
btmonier, line 31: This is kind of confusing. semantically, you probably do not need an inverse flag at all. Since the default is false, this could be enough:

``` kotlin
val bySample by option(
    "--by-sample",
    help = "Create a report with a read count for each sample by reference range. " +
        "The default is a report with only one row per reference range. " +
        "--by-sample report does not use --target-sample-name."
).flag()
```

src/main/kotlin/net/maizegenetics/phgv2/pathing/ReadMappingCountQc.kt
```diff
@@ -20,18 +21,22 @@ class ReadMappingCountQc : CliktCommand("Read mapping count QC") {
     val readMappingFile by option(help = "Read mapping file")
         .required()
 
-    val targetSampleName by option(help = "Target sample name")
+    val targetSampleName by option(help = "Target sample name. Required but not used when the --by-sample flag is set.")
```
btmonier: In relation to `--by-sample`, when `targetSampleName` is declared `.required()`, but the help text on `--by-sample` says "does not use `--target-sample-name`". So a user who only wants the by-sample report is forced to pass a placeholder value (e.g., `--target-sample-name=foo`). You could make this nullable and validate it post-parsing. For example:

``` kotlin
val targetSampleName by option(help = "Target sample name. Required unless --by-sample is set.")

override fun run() {
    logCommand(this)
    if (!bySample && targetSampleName.isNullOrBlank()) {
        throw UsageError("--target-sample-name is required unless --by-sample is set")
    }
    processReadMappingCounts(hvcfDir, readMappingFile, targetSampleName ?: "", outputDir, bySample)
}
```

src/main/kotlin/net/maizegenetics/phgv2/pathing/ReadMappingCountQc.kt
```diff
@@ -43,9 +48,63 @@ class ReadMappingCountQc : CliktCommand("Read mapping count QC") {
...
+    fun writeDetailReport(hapIdCounts: Map<String, Int>,
+                          outputFile: String,
+                          rangeToHapIds: Map<ReferenceRange,List<String>>,
+                          hapIdToSampleGamete: Map<String, List<SampleGamete>>,
+                          referenceRangeCounts: Map<ReferenceRange, Int>) {
+
+        val referenceRangeList = referenceRangeCounts.keys.toList().sorted()
+
+        bufferedWriter(outputFile).use { writer ->
+            writer.write("RefRange\tSampleName\tHapID\tHapCount\tTotalCount \n")
```
btmonier: I don't think this is intentional but there is a whitespace that should be removed after `TotalCount` - this could affect other parsers downstream.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ReadMappingCountQc.kt
```diff
@@ -43,9 +48,63 @@ class ReadMappingCountQc : CliktCommand("Read mapping count QC") {
...
+            referenceRangeList.forEach { refRange ->
+                val hapids = rangeToHapIds[refRange]
+                if (hapids != null) {
+                    val totalCount = referenceRangeCounts[refRange]
+                    val hapidSamplePairs = hapids.flatMap {hapid -> hapIdToSampleGamete[hapid]!!.map { sample -> Pair(hapid, sample)}}
```
btmonier, line 82: This could be written more safe to avoid a possible null pointer exception:

``` kotlin
val hapidSamplePairs = hapids.flatMap { hapid ->
    hapIdToSampleGamete[hapid]
        ?.map { sample -> hapid to sample }
        ?: emptyList()
}
```

btmonier (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ReadMappingCountQc.kt
pjbradbury: Clickt documentation of the flag option suggests including the inverse. But I agree with you and think that does not really apply here. Happy to get rid of it.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ReadMappingCountQc.kt
pjbradbury: Good suggestion.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ReadMappingCountQc.kt
pjbradbury: Definitely not intentional. I will fix it.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ReadMappingCountQc.kt
pjbradbury: Since every hapid has to be associated with at least one sample gamete, this should never happen. If it does, throwing an error would be appropriate. I am going to leave it as is.

aberthel (approved):

zrm22 (approved):

PR #365, opened 2026-05-14
Reads mapping multiple ranges

pjbradbury:
## Description

This pull request modifies MapReads by adding a new parameter, --min-single-range. The values for this parameter can range from 0.0 to 1.0. It is the minimum proportion of mapping hits as reported by ropebwt3 that hit a single range. If one, all hits must come from one range. If 0, then there is no restriction on the minimum number of hits coming from a single range. The value defaults to zero, in which case the is no change to the mapping results.

## Type of change

## Checklist:

## Changelog entry

Adds --min-single-range parameter to map-reads, which for each read, determines the minimum proportion of hits that must come from a single reference range.

```release-note

```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReads.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReads.kt
@@ -70,7 +70,7 @@ class MapReads : CliktCommand(help="Map reads to a pangenome using ropeBWT3") {
         .int()
         .default(70)
 
-    val minSingleRange by option(help = "Minimum proportion of read mappings to a single range. " +
+    val minSingleRange by option(help = "Minimum proportion of reads mapping to a single range. " +
             "By default reads mapping to more than one range will not be used. Value must be between 0.0 and 1.0. (Default = 1.0)")
         .double()
         .default(1.0)
--- a/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReads.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReads.kt
@@ -9,6 +9,8 @@ import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.default
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.options.validate
+import com.github.ajalt.clikt.parameters.types.double
 import com.github.ajalt.clikt.parameters.types.int
 import net.maizegenetics.phgv2.api.HaplotypeGraph
 import net.maizegenetics.phgv2.api.ReferenceRange
@@ -19,6 +21,7 @@ import net.maizegenetics.phgv2.pathing.ReadInputFile
 import org.apache.logging.log4j.LogManager
 import java.io.BufferedReader
 import java.io.InputStreamReader
+import kotlin.collections.contains
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReads.kt
```diff
@@ -199,38 +207,38 @@ class MapReads : CliktCommand(help="Map reads to a pangenome using ropeBWT3") {
...
+            //matches to multiple reference ranges
+            val refRangeCounts = bestHapIds.mapNotNull { hapIdToRefRangeMap[it] }
+                .flatten()
+                .groupingBy { it }
+                .eachCount()
+            val maxRefRangeCount = refRangeCounts.maxOf { it.value }
+            val totalRefRangeCount = refRangeCounts.values.sum()
+
+            //control how many hits ar allowed to multiple reference ranges
+            if (maxRefRangeCount / totalRefRangeCount >= minSingleRange) {
```
btmonier: Should `maxRefRangeCount` be converted to a double to prevent integer division?

src/test/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReadsTest.kt
```diff
@@ -79,6 +80,12 @@ class MapReadsTest {
         assertEquals(1, noHvcfDir.statusCode)
         assertEquals("Usage: map-reads [<options>]\n\n" +
                 "Error: missing option --hvcf-dir\n", noHvcfDir.stderr)
+
+        val outOfRange = mapReads.test("--min-single-range 2.0 --index testIndex --read-files test1.fq --output-dir testDir --hvcf-dir ${TestExtension.smallSeqInputDir}")
+        assertEquals(1, noOutputDir.statusCode)
```
btmonier: This looks like it was copied from above. Should this be `outOfRange.statusCode` instead?

btmonier (approved):

aberthel (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReads.kt
```diff
@@ -80,26 +88,26 @@ class MapReads : CliktCommand(help="Map reads to a pangenome using ropeBWT3") {
...
+        mapAllReadFiles(index, readInputFiles.getReadFiles(), outputDir, threads, minMemLength, maxNumHits, condaEnvPrefix, hapIdToRefRangeMap, maxStart, minEnd, minSingleRange)
     }
 
     /**
      * Function to map all the read files in the keyFileDataEntries to the index
      */
     fun mapAllReadFiles(index: String, keyFileDataEntries: List<KeyFileData>, outputDir: String, threads: Int,
                         minMemLength: Int, maxNumHits: Int, condaEnvPrefix: String,
-                        hapIdToRefRangeMap : Map<String, List<ReferenceRange>>, maxStart: Int, minEnd: Int) {
+                        hapIdToRefRangeMap : Map<String, List<ReferenceRange>>, maxStart: Int, minEnd: Int, minSingleRange: Double) {
```
zrm22: Should each of these functions passing in the minSingleRange have a default value?  Just in case the function gets called externally?

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReads.kt
pjbradbury: Yes. Will do that

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReads.kt
pjbradbury: Good idea. That will allow me to eliminate a couple of changes to testing code.

src/test/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/MapReadsTest.kt
pjbradbury: yes. i will change that

PR #371, opened 2026-07-22
Fix spline alt allele issue

zrm22:
## Description

Fixes a number of issues with the splines but also adds in an alternative way to do ropebwt indexing.

## Type of change

## Checklist:

## Changelog entry

```release-note
Fixing some bugs related to spline mapping.
```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ConvertRopebwt2Ps4gFile.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ConvertRopebwt2Ps4gFile.kt
@@ -58,6 +58,8 @@ sealed class BedInputFile {
 }
 
 
+data class ContigAndGamete(val contig: String, val gamete: String )
+
 /**
  * This class will convert a RopebwtBed file to a PS4G file.  It will only work with ropebwt3 files where the reads are
  * aligned to the whole assembly chromosomes using the mem command.  MEMs are Maximal Exact Matches and are used to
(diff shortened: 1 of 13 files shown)


pjbradbury (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ConvertRopebwt2Ps4gFile.kt
```diff
@@ -121,8 +128,19 @@ class ConvertRopebwt2Ps4gFile : CliktCommand(help = "Convert RopebwtBed to PS4G"
         //check to make sure its not empty and the files exist
         check(bedFilesToProcess.isNotEmpty()) {"No Ropebwt Bed files provided"}
         bedFilesToProcess.forEach { ropebwtBed ->
-            processSingleRopeBwtBed(ropebwtBed, outputDir,splineLookup, gameteIndexMap, sampleGameteIndexMap, command,minMemLength, maxNumHits, maxRange, sortPositions)
+            processSingleRopeBwtBed(ropebwtBed, outputDir,splineLookup, gameteIndexMap, sampleGameteIndexMap, command,minMemLength, maxNumHits, maxRange, bedSampleNameToChrAndSample, sortPositions )
+        }
+    }
+
+    fun buildSampleNameSplitMap(chrIndexMap: Map<String, Int>, gameteIndexMap: Map<String, Int>, sampleNameFirst: Boolean = false): Map<String, ContigAndGamete> {
```
btmonier, line 135: Does `chrIndexMap` hold assembly contig names here, or only reference ones? I traced it as only getting `refChr`/`chrom`, and the BED names look like they're the assembly contigs the spline knots are keyed on. I am wondering whether a scaffold or a different named assembly contig would miss the map and get dropped by the new `isEmpty()` filter. Would building this from the spline knot keys instead sidestep it?

btmonier (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ConvertRopebwt2Ps4gFile.kt
zrm22: Yes it looks like you are right that it only picks up the ref chromosomes.  They would likely get dropped in this case.  I will add this to another task as we really need to get the rest of this PR in(mainly the chr_sample swapping for the indexing).  The whole spline thing is likely going to be deprecated very soon as the new ref_map commands work much better(and now we have it working so it can handle 1 bp gaps in the MEMs as well).

aberthel (approved):

PR #372, opened 2026-07-29
Forward backward hmm

pjbradbury:
## Description

This pull request adds the class ImputeBinProbabilities, which imputes genotype probabilities for each state at each position/bin. It takes a Ps4g file as input and generates a list of genotypes and probabilities by bp position. To reduce the volume of output, it only writes out probabilities greater than 0.05 * the maximum probability at a position. That ratio can be changed with one of the method parameters.

## Type of change

## Checklist:

## Changelog entry

Adds impute-bin-probabilities command

```release-note
This adds the phg command impute-bin-probabilities. Impute-bin-probabilities calculates the probability of each genotype (state) for each bin in a Ps4g file. It runs the HMM forward-backward algorithm to calculate those values. The backward-forward algorithm is a useful addition to the Viterbi algorithm because it can help explain the choices made during path finding and indicate the level of confidence for the genotype calls.
```


src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/AbstractEmissionProbability.kt
```diff
@@ -0,0 +1,115 @@
...
+
+        //get probabilities for all ordered pairs in the parentList
+        val probabilityArray  = DoubleArray(numberOfPairs)
+        var arrayPtr = 0
+        if (numberOfReads > maxNumberOfTrials) {
+            val distribution = BinomialDistribution(numberOfReads, probCorrect)
+            for (parent1 in parentList) {
+                for (parent2 in parentList) {
+                    probabilityArray[arrayPtr++] = if (parent1 == parent2) {
+                        val indexClassCounts = indexCountsForOneIndex(gameteSetList, parent1)
```
zrm22, line 62: I wonder if there is a way to refactor this a bit.  Maybe pull the logic out to a function and use this here and for the else loops as well.

zrm22 (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/AbstractEmissionProbability.kt
pjbradbury: That's a reasonable suggestion, but I am  looking at changing how the diploid emission probability is calculated as a next project so won't address that as part of this pr.

docs/imputation_ml.md
```diff
@@ -70,6 +70,20 @@ imputation using the PHG:
...
+* Impute parental genotypes from a PS4G file using the Viterbi algorithm:
+  ```shell
+  phg impute-path-from-ps4g \
+      --read-file /path/to/ps4g_file.ps4g \
+      --out-path-dir /my/output_dir
+  ```
+
+* Calculate state probabilities from a PS4G file using the forward-backward algorithm:
+  ```shell
+  phg impute-path-from-ps4g \
```
btmonier: Shouldn't this be the new `impute-bin-probabilities` clikt command?

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputeBinProbabilities.kt
```diff
@@ -0,0 +1,266 @@
...
+ * Input is one or more PS4G files produced by `align-reads`, supplied either through a tab-delimited
+ * key file (`--path-keyfile`) or directly (`--read-file`). For each sample, a per-contig HMM is
+ * built from:
+ *  - an initial state distribution derived from [inbreedCoef] and the number of parents,
+ *  - a transition matrix parameterized by [probSame] (haploid) or [DiploidTransitionProbability] (diploid),
+ *  - emission probabilities from [EmissionProbabilityForForwardBackward] driven by [probCorrect].
+ *
+ * Results are written per sample to `<outputDir>/<sampleName>_imputed_probabilities.txt`.
+ */
+class ImputeBinProbabilities: CliktCommand(help = "Impute best haplotypes from a Ps4g file.")  {
```
btmonier: This help string is identical to `ImputePathFromPs4g`'s, so `phg --help` will list the
two commands with the same one-line description. Maybe something like:

 > Calculate posterior haplotype probabilities from a Ps4g file using forward-backward 

to distinguish it from the Viterbi command?

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputeBinProbabilities.kt
```diff
@@ -0,0 +1,266 @@
...
+            "--path-keyfile", help = "Name of tab-delimited key file.  Columns for samplename and filename" +
+                    " are required. Files must be ps4g files created by align-reads. A value must be entered for " +
+                    "either --path-keyfile or --read-files."
+        )
+            .convert { PathInputFile.KeyFile(it) },
+        option("--read-file", help = "The name of a ps4g file created by align-reads.")
+            .convert { PathInputFile.ReadFiles(it) }
+    ).single().required()
+
+    val outputDir by option(
```
btmonier, line 56: A couple of things:
* The help text looks like it came over from ImputePathFromPs4g. This command writes
`<sampleName>_imputed_probabilities.txt,` and there aren't half-open coordinates in the
output, just a bin start.

* Separately, and more of an open question: this is `--output-dir` while the sibling
command uses `--out-path-dir`. Is there a reason to keep them different? Since this
command hasn't shipped yet, now is the cheap moment to align them if we want to,
though I don't know whether `--out-path-dir` is load-bearing in anyone's scripts.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputeBinProbabilities.kt
```diff
@@ -0,0 +1,266 @@
...
+    )
+        .required()
+
+    val imputeType by option(
+        help = "The type of genotype probabilities to calculate. Must be lower case 'haploid' or 'diploid' (without quotes)."
+    )
+        .choice("haploid", "diploid")
+        .default("haploid")
+
+    val probCorrect by option(help = "The probability that a read maps to correct haplotype. Default = 0.98")
```
btmonier, line 68: Probably worth adding [`.validate {}`](https://ajalt.github.io/clikt/api/clikt/com.github.ajalt.clikt.parameters.options/validate.html) to `probCorrect`, `probSame`, and `inbreedCoef` 

`FindPaths` does this for the same three (lines 117-143): 

* `prob-correct` in `0.5..1.0`,
* `prob-same-gamete` in `0.01..1.0`, 
* `inbreeding-coefficient` in `0.0..1.0`. 

the help text here already promises `0.0..1.0` for `inbreed-coef`, so it would just be making that enforceable, and out of range values would fail with a usage error instead of producing quietly
wrong posteriors

btmonier (approved):

aberthel (approved):

docs/imputation_ml.md
pjbradbury: yes, it should.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputeBinProbabilities.kt
pjbradbury: Changed

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputeBinProbabilities.kt
pjbradbury: Changed the help text for outputDir. I am going to leave it outputDir. I would change ImputePath to match but some  users may already have scripts with the other, so probably best to leave as is.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputeBinProbabilities.kt
pjbradbury: Good idea! It means I will need to add some additional checks in the unit test.

PR #373, opened 2026-08-04
Add contig list to restrict contigs to be imputed from ps4g files

pjbradbury:
## Description

The code for ImputePathFromPs4g and ImputeBinProbabilities did not impute paths for any contigs starting with "scaf". This made sense for many maize data sets but may not always be desirable. Instead, this branch removes the contig != "scaf" filter and provides options for the user to supply a contig list. The branch also updates unit tests to reflect these changes, adds some new tests to cover the new options, and updates the command documentation.

## Type of change

## Checklist:

## Changelog entry

Adds the option for users to supply a list of contigs to be imputed to ps4g imputation.
```release-note

```

### DIFF
--- a/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputePathFromPs4g.kt
+++ b/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputePathFromPs4g.kt
@@ -1,6 +1,7 @@
 package net.maizegenetics.phgv2.pathing.ropebwt
 
 import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.UsageError
 import com.github.ajalt.clikt.parameters.groups.mutuallyExclusiveOptions
 import com.github.ajalt.clikt.parameters.groups.required
 import com.github.ajalt.clikt.parameters.groups.single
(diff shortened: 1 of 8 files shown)


src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputePathFromPs4g.kt
```diff
@@ -17,15 +17,40 @@ import net.maizegenetics.phgv2.pathing.MostLikelyPs4gParents
...
+         * If [contigString] names an existing file, the contigs are read from it, one per line.
+         * Otherwise the value is treated as a comma-separated list of contig names. A blank value
+         * returns an empty set, which [Ps4gFileReader] takes to mean "use every contig in the file".
+         */
+        fun buildContigSet(contigString: String): Set<String> {
+            return if (contigString.isNotBlank()) {
+                try {
+                    val filePath = Path.of(contigString)
+                    if (filePath.exists()) getBufferedReader(filePath.toFile()).use { it.readLines().toSet() }
+                    else contigString.split(",").toSet()
```
btmonier: Could we trim the contig names and filter out blank entries here? For example, `"chr1, chr2"` would currently treat the second value as `" chr2"` and therefore fail to match

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputePathFromPs4g.kt
```diff
@@ -17,15 +17,40 @@ import net.maizegenetics.phgv2.pathing.MostLikelyPs4gParents
...
+         *
+         * If [contigString] names an existing file, the contigs are read from it, one per line.
+         * Otherwise the value is treated as a comma-separated list of contig names. A blank value
+         * returns an empty set, which [Ps4gFileReader] takes to mean "use every contig in the file".
+         */
+        fun buildContigSet(contigString: String): Set<String> {
+            return if (contigString.isNotBlank()) {
+                try {
+                    val filePath = Path.of(contigString)
+                    if (filePath.exists()) getBufferedReader(filePath.toFile()).use { it.readLines().toSet() }
```
btmonier: Possible edge case - what behavior do we want when `--contigs-to-use` points to an empty file? Since that produces an empty set, would `Ps4gFileReader` (line 51) interpret it the same as not providing a contig filter and process every contig?

btmonier (approved):

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputePathFromPs4g.kt
pjbradbury: Good idea.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/ImputePathFromPs4g.kt
pjbradbury: Seems unlikely but I added a check for it

zrm22 (approved):

PR #375, opened 2026-09-17
Ps4g emission improvements

pjbradbury:
## Description

This changes the way that emission probability is calculated for the Viterbi algorithm and greatly improves the efficiency of the algorithm by reducing the number of actual transition calculations. It uses the same basic method for both haploid and diploid emission, just forcing inbreeding coefficient = 1 for haploid. Testing with simulated data confirmed that accuracy is equal to or better than the previous algorithm and about 100x faster for diploid pathing. It changes the default for --prob-same to 0.999999999 from 0.9999 and --n-parents now applies to both haploid and diploid path finding. 

## Type of change

- [x ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x ] I have updated relevant documentation

## Changelog entry

```release-note
Improves accuracy and speed of path finding for impute-path-from-ps4g
```


src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/BuildPresenceTable.kt
```diff
@@ -0,0 +1,293 @@
...
+        return counts
+    }
+
+    fun writeTable(out: File, window: Int, taxa: List<String>, chroms: List<ChromBlock>) {
+        DataOutputStream(out.outputStream().buffered(1 shl 20)).use { output ->
+            output.write("PAVPRES".toByteArray(Charsets.ISO_8859_1))
+            output.write(1)
+            fun int(value: Int) = output.writeInt(Integer.reverseBytes(value))
+            fun str(value: String) = value.toByteArray(Charsets.UTF_8).let { int(it.size); output.write(it) }
+            int(window)
```
zrm22, line 236: maybe rename these functions, at first glance they looked like conversion functions instead of writers

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/BuildPresenceTable.kt
```diff
@@ -0,0 +1,293 @@
...
+            int(taxa.size)
+            taxa.forEach { str(it) }
+            int(chroms.size)
+            chroms.forEach { str(it.name); int(it.nWindows) }
```
zrm22, line 240: put these in the for loop below.  Less iterations over the chroms list.

zrm22: Actually this might break things in the output format.  It would be nice to walk once over the list

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/BuildPresenceTable.kt
```diff
@@ -0,0 +1,293 @@
...
+            while (needed < size + other.size) needed *= 2
+            if (needed != values.size) values = values.copyOf(needed)
+            other.copyInto(values, size)
+            size += other.size
+        }
+
+        /** Sorted copy with duplicates removed. */
+        fun sortedDistinct(): LongArray {
+            if (size == 0) return LongArray(0)
+            val copy = values.copyOf(size)
```
zrm22, line 281: Maybe go to sorted set then back to LongArray?  Would simplify the logic a lot, but might be slow due to auto-boxing.  Worth a test.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/GameteSetEmissionProbability.kt
```diff
@@ -0,0 +1,225 @@
...
+    val pavThreshold: Double = 0.02,
+    val pavDamping: Double = 1.0
+) {
+    val parentList = parentSet.sorted()
+    val nParents = parentList.size
+    val positionList = readMap.keys.sorted()
+
+    private val myLogger = LogManager.getLogger(GameteSetEmissionProbability::class.java)
+
+    private val parentToLocal = HashMap<Int, Int>(nParents * 2).also { map ->
```
zrm22, line 93: There might be a more kotliny way to do this.  
`parentList.mapIndexed{} ...`

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/BuildPresenceTable.kt
```diff
@@ -0,0 +1,293 @@
...
+        val names = ArrayList<String>()
+        val seqLengths = ArrayList<Long>()
+        getBufferedReader(lengths).forEachLine { line ->
+            if (line.isBlank()) return@forEachLine
+            val parts = line.split('\t')
+            names.add(parts[0])
+            seqLengths.add(parts[1].trim().toLong())
+        }
+
+        val taxa = names.map { it.substringBefore("_") }.distinct().sorted()
```
btmonier, line 124: `build-presence-table` seems to assume the index was built with `--sample-name-first`, and with the default `contig_sample` naming it looks like it would write an empty table without complaining. Should the help/docs call out that requirement, and would a `require(refChroms.isNotEmpty())` (or similar) be reasonable so a wrong `--reference-prefix` or naming order fails loudly?

docs/imputation_ml.md
```diff
@@ -281,31 +281,62 @@ phg impute-path-from-ps4g \
     --read-file /data/my_ps4g_file.ps4g \
     --output-dir /results/ps4g \
     --path-type diploid \
-    --inbreed-coef 0.5 \
-    --n-parents 10 \
-    --bin-size 5000
+    --n-parents 15 \
```
btmonier, line 284: A few doc spots look like they predate the later commits: 
* `--n-parents` still says diploid-only
* the `--prob-same` default in the table is `0.9999`, 
* the `--window-size 50_000` example won't parse with Clikt since this gets taken in as a string and `String.toInt()` does not treat `_` as a digit separator
* `PresenceTable` KDoc still points at the Python script.

btmonier (approved):

zrm22 (approved):

docs/imputation_ml.md
pjbradbury: Fixed these except I could not find where a kdoc points to a PresenceTable python script.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/BuildPresenceTable.kt
pjbradbury: Good catch. Its not hard to check whether the sample name is first or last and handle both cases. So I think that's preferable. Also, I see that the code assumes chromosome names start with "chr", which will eventually cause problems since there do not seem to be any chromosome naming standards across species.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/BuildPresenceTable.kt
pjbradbury: I am going to leave this alone. It runs fast and only gets done once per lift file.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/GameteSetEmissionProbability.kt
pjbradbury: Agree. The original is a bit convoluted.

src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/BuildPresenceTable.kt
pjbradbury: I am inclined to leave it this way. The method comments say that it is written specifically to avoid boxed arrays because of memory requirements.

docs/imputation_ml.md
btmonier: Nothing major, but something I saw in the comment section [here](https://github.com/maize-genetics/phg_v2/blob/308b17271812b2c6a7c7c981642ef7193dec6a15/src/main/kotlin/net/maizegenetics/phgv2/pathing/ropebwt/PresenceTable.kt#L26-L31) - there is a reference to a prior Python script.

aberthel (approved):

